PR #41, opened 2026-06-01
fix(checkpoint): re-prompt setup when the default snapshot is stale

madarco:
## Problem

Running `agentbox claude` in a project whose `box.defaultCheckpoint` points at a **stale** snapshot announced:

```
●  starting from checkpoint "setup"; skipping agentbox.yaml setup
◐  [image] #22 RUN curl -fsSL https://claude.ai/install.sh | bash -s stable
```

i.e. it skipped setup *and* rebuilt the base image — then booted from the old checkpoint layers anyway. The wizard's skip decision only checked that the manifest **existed**, ignoring whether the checkpoint was still valid.

Repro case: a schema-2 (pre-versioning) `setup` checkpoint while the base image had been re-prepared since.

## Fix

- **`evaluateCheckpoint` (`apps/cli/src/checkpoint-lookup.ts`)** — classifies a ref as `fresh` / `stale` / `missing`:
  - docker: orphaned image → missing; schema-2/no-fingerprint or fingerprint mismatch → stale
  - cloud: dead snapshot → missing; legacy schema-1 or fingerprint mismatch → stale
- **Wizard (`apps/cli/src/wizard.ts`)** now branches on that:
  - **interactive + stale default** → re-prompt "snapshot is stale — recreate or use anyway?"; recreate discards the ref and re-runs setup
  - **`-y` / non-TTY / autolaunch** → treat `-y` as *yes to the wizard*: `claude -y` runs setup, `create -y` makes a bare box; a stale/dead default is discarded so the box provisions from the current base (a warm checkpoint behind an existing `agentbox.yaml` is kept, not cold-rebuilt)
  - **explicit `--snapshot`** is always honored as-is (the create path still logs its loud staleness warning)
- **`discardCheckpoint`** signal threads through `claude.ts` / `create.ts` so the ref is dropped before `createBox` / `provider.create`.
- **Cloud parity** — `CloudCheckpointManifest` bumped to schema 2 with `baseProvider` / `baseFingerprint` / `cliVersion`, captured at checkpoint-create time, so cloud staleness is verifiable going forward.

## Verification

- `pnpm typecheck` 21/21, `pnpm lint` clean, build green.
- Evaluator checked against real on-disk data: stale-by-schema-2, stale-by-fingerprint-mismatch, and missing-manifest all classified correctly.
- `claude -y` end-to-end on the stale project: no "starting from checkpoint", discarded → built base → provisioned a fresh git-worktree box, came up clean.

## Note

The interactive recreate path re-runs setup but does not auto-recapture the `setup` checkpoint (recapturing stays a manual `agentbox checkpoint create setup`, same as today).

---

> [!NOTE]
> **Medium Risk**
> Changes first-run and default-checkpoint boot behavior for create/claude across Docker and cloud; explicit `--snapshot` is preserved but default-checkpoint and non-interactive paths now drop or re-prompt on staleness.
> 
> **Overview**
> Fixes misleading **“starting from checkpoint …; skipping setup”** when `box.defaultCheckpoint` still exists but is **orphaned**, **dead on the cloud backend**, or **captured against an older base** than the current prepared image/snapshot.
> 
> **`evaluateCheckpoint`** replaces the old existence-only check with **`fresh` / `stale` / `missing`**: Docker verifies the backing image and compares `baseFingerprint` to prepared context; cloud probes snapshot liveness (pruning dead manifests) and compares fingerprints via new **`currentCloudBaseFingerprint`**.
> 
> The **setup wizard** uses that classification: **fresh** (or explicit **`--snapshot`**) skips setup as before; **stale defaults** get an interactive **recreate vs use anyway** prompt; **missing/orphaned defaults** are dropped. **`discardCheckpoint`** flows into **`agentbox create`** / **`agentbox claude`** so provisioning uses the current base instead of the bad ref. **`-y` / non-TTY** paths discard stale/missing defaults when there is no `agentbox.yaml`, while keeping a stale warm start when config already exists.
> 
> **Cloud checkpoints** bump manifests to **schema 2** with `baseProvider`, `baseFingerprint`, and `cliVersion` at capture time so staleness is verifiable going forward (legacy schema 1 treated as unverifiable/stale).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f299d0131d818d184629feeceada3b1a9cfde935. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/checkpoint-lookup.ts
+++ b/apps/cli/src/checkpoint-lookup.ts
@@ -1,47 +1,123 @@
 /**
- * Provider-aware checkpoint existence check used by the wizard. The default
+ * Provider-aware checkpoint evaluation used by the wizard. The default
  * checkpoint name lives in a single config field (`box.defaultCheckpoint`),
- * but the actual artifact may exist for Docker, for Daytona, both, or
- * neither. The wizard consults this helper before announcing "starting from
- * checkpoint …" — if the named checkpoint doesn't exist for the active
- * provider, the wizard falls through to the normal setup flow instead of
- * misleadingly skipping it.
+ * but the actual artifact may exist for Docker, for a cloud backend, both, or
+ * neither — and even when it exists it may be *stale* (captured against a base
+ * image/snapshot that has since been rebuilt) or *orphaned* (its underlying
+ * Docker image / cloud snapshot is gone).
  *
- * For cloud providers the local manifest is only half the story: the provider
- * snapshot it points at can expire or be deleted out-of-band. We probe the
- * backend for liveness and prune the dangling manifest when it's gone, so the
- * wizard re-asks the setup wizard (start-from-scratch) rather than booting
- * from a snapshot that would 410 mid-create.
+ * `evaluateCheckpoint` collapses all of that into three states so the wizard
+ * can decide whether to silently skip setup (`fresh`), re-prompt the user to
+ * recreate it (`stale`), or fall through to normal setup (`missing`). Without
+ * this, a stale checkpoint would announce "starting from checkpoint …; skipping
+ * setup" and then quietly rebuild the base image while booting from the old
+ * layers — the exact confusion this module exists to prevent.
  */
 
 import type { ProviderName } from '@agentbox/core';
-import { resolveCheckpoint } from '@agentbox/sandbox-docker';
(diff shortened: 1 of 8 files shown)


madarco:
bugbot run

PR #66, opened 2026-06-05
fix(wizard): prompt on a stale default checkpoint even for configured projects

madarco:
## Problem
The interactive stale-snapshot prompt — *"Snapshot X is stale … Recreate it from the current base? (No = use it anyway)"* (`wizard.ts`) — sat **after** the `hasAgentboxYaml` early-return. So any project with an `agentbox.yaml` (i.e. every configured project) skipped the prompt entirely and silently booted the old base layers, missing base-image updates (e.g. a CLI upgrade / new in-box behavior).

The only remaining signal was the `createBox` log warning, but the CLI pipes that channel into the clack **spinner** (`s.message`), so the multi-line WARNING is immediately overwritten and scrolls by unseen. Net: a configured project got **neither** a visible prompt nor a visible warning when restoring a stale default checkpoint.

This was found via a real case: an `optima` box kept booting from a checkpoint frozen with the old `agentbox-ctl` (no new scraper), with no visible warning.

## Fix
Hoist the stale-default-checkpoint prompt **above** the configured-project early-return — mirroring the stale-**cloud-base** prompt, which already deliberately runs there (its comment: *"re-prompt BEFORE the agentbox.yaml early-return"*). Checkpoint staleness is about base freshness, not whether the project is set up.

- On **recreate**: discard the stale checkpoint and fall through to the setup flow, which recaptures a fresh `--set-default` snapshot on the current base. Fast when an `agentbox.yaml` already exists (the `/agentbox-setup` skill ends with `agentbox checkpoint --set-default`). The `hasAgentboxYaml` early-return now fires only when **not** recreating.
- On **use it anyway**: keep the warm checkpoint, exactly as before.
- Non-interactive (`-y`/`-i`) behavior is unchanged.

## Verification
- `pnpm build` / typecheck / lint clean; wizard + commands unit tests pass (32).
- **PTY harness against the real `optima` project** (has `agentbox.yaml` + a stale default checkpoint): `agentbox claude` now shows the recreate prompt —
  > Snapshot "optima-bbce43c9a-1" is stale (base image updated since capture (captured e601c80c5cfa, current f995e4c91767)). Recreate it from the current base? (No = use it anyway)

  — which it did **not** show before this change. Aborted at the prompt so no box was provisioned.

---

> [!NOTE]
> **Low Risk**
> Localized CLI wizard and env-sentinel changes around checkpoint staleness; behavior is covered by new unit tests with no auth or data-path impact.
> 
> **Overview**
> **Moves the stale default-checkpoint prompt ahead of the configured-project (`agentbox.yaml`) short-circuit**, so projects that already have config still see *"Snapshot … is stale … Start from base and run Setup Wizard?"* instead of silently booting old base layers. Choosing **recreate** discards the stale snapshot and continues into setup (which recaptures a fresh default on the current base); **use it anyway** is unchanged.
> 
> **Fixes the Docker `agentbox create` → `agentbox claude` handoff** when the user chose recreate: the inner autolaunch pass no longer uses `nonInteractiveOutcome`, which kept a stale default checkpoint for configured projects. A new `WIZARD_RECREATE_ENV` sentinel (and `recreate` on `WizardOutcome`) forces `launch-with-prompt` with `discardCheckpoint: true` on the inner pass.
> 
> Unit tests cover the autolaunch recreate path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd2b2596af06ba56e253bdf2fe7fec8ed62583b7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/create.ts
+++ b/apps/cli/src/commands/create.ts
@@ -32,6 +32,7 @@ import {
   serializeEnvFilesForEnv,
   WIZARD_AUTOLAUNCH_ENV,
   WIZARD_ENV_FILES_ENV,
+  WIZARD_RECREATE_ENV,
 } from '../wizard.js';
 import { evaluateBaseFreshness } from '../checkpoint-lookup.js';
 import { runPrepare } from './prepare.js';
@@ -336,8 +337,11 @@ export const createCommand = new Command('create')
     }
     // Drop a stale/dead default checkpoint so the box provisions from the
     // current base. On the docker switch-to-claude re-dispatch below the
-    // default isn't forwarded as `--snapshot`, so the inner `agentbox claude`
-    // re-evaluates and discards it too — no extra signal needed.
+    // default isn't forwarded as `--snapshot`; the inner `agentbox claude`
+    // re-evaluates a *missing/dead* default and discards it too. A *stale*
+    // default that the user chose to RECREATE is forwarded explicitly via
+    // WIZARD_RECREATE_ENV, because the inner non-interactive pass would
+    // otherwise keep a stale checkpoint for a configured project.
     const effectiveCheckpointRef = wiz.discardCheckpoint ? undefined : checkpointRef;
     if (wiz.action === 'switch-to-claude' && isDocker) {
       // Docker: hand off to `agentbox claude` whose default action creates +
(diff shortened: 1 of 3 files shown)


madarco:
bugbot run

PR #67, opened 2026-06-05
feat(cp): stream tar, add --exclude, block oversized copies with a tree

madarco:
## Problem

`agentbox cp` (and the `carry:` copy step) buffered the whole tarball into a Node Buffer via execa's `encoding: 'buffer'`. execa's default `maxBuffer` is **100 MB**, so any folder larger than that silently failed with *"tar: Write error"* — hit live carrying a 2.4 GB folder into a box (`optima-b2c79b76e`).

## Changes

- **Streaming fix (root cause):** new `streamTarPipe` streams host `tar` ⇄ `docker exec tar` with stdout buffering disabled, so copies are unbounded and memory stays flat. Applied to cp upload/download, the carry-dir path, and cloud cp.
  - First attempt with execa's `.pipe()` still failed: each `execa()` starts buffering its stdout at creation *before* `.pipe()` redirects it. Fixed with `buffer: { stdout: false }` + raw stream piping.
- **`--exclude` + auto-skip defaults:** `--exclude=<glob|name>` (repeatable) and `--no-default-excludes`. Heavy regenerable dirs (`.git`, `node_modules`, `bin`, `obj`, `packages`, `dist`, `.next`, `target`) drop by default. Threaded through `agentbox-ctl cp` → relay `cp.*` RPC → host `cp`, plus a per-item carry `exclude:`.
- **Block oversized uploads:** over `box.cpMaxBytes` (new config key, default 100 MiB, measured *after* excludes) the copy is refused unless `--yes`, printing a recursive `du`-style tree of the biggest folders/subfolders + a strategy (split per-folder / `--exclude` / `--yes`). New `dir-breakdown.ts` (exclude-aware size + tree) with unit tests.
- carry-resolve sizing + `carrySourceHash` are exclude-aware so the cap and resync track only what actually copies.
- Docs: `cli.mdx`, `agentbox-yaml.mdx`, `features.md`; `agentbox-setup` + `agentbox-info` skills.

## Verification (live, box `optima-b2c79b76e`)

- **Block + tree** on the 2.4 GB folder → 408 MB after default excludes (dropped `.git` + `packages`), still blocked, full nested tree + concrete split commands.
- **Streaming**: 440 MB copied with `--yes` in ~5s (old code died at 100 MB), vscode-owned, no `.git`/`packages`.
- `--exclude=Application` brought it under the limit (78 MB, no `--yes` needed); small-file copy + box→host download both fine.
- Full suite green (470 CLI tests incl. 11 new), typecheck + lint clean.

## Note for review

Per the approved plan, the carry block also auto-applies the default excludes (not just cp) — a `carry:` of a source dir now silently drops `node_modules`/`.git`/etc. Almost always desired (regenerated by install), but it does change carry semantics; flag if it should honor only explicit `exclude:`.

---

> [!NOTE]
> **Medium Risk**
> Changes default carry semantics (heavy dirs now dropped silently) and touches all copy paths (docker, cloud, relay); streaming fix is low-risk but behavior shifts may surprise users who relied on copying `.git`/`node_modules` via carry.
> 
> **Overview**
> Fixes silent copy failures on large host→box transfers by **streaming** tar pipes (`streamTarPipe` with `buffer: { stdout: false }`) instead of buffering whole archives through execa’s ~100 MB limit; applies to `agentbox cp`, `carry:` directory copies, and cloud cp.
> 
> Adds **`--exclude`**, **`--no-default-excludes`**, and default skipping of heavy dirs (`.git`, `node_modules`, build outputs, etc.), wired through Docker/cloud providers, **`carry:` `exclude:`**, and **`agentbox-ctl cp`** → relay RPC. Uploads over **`box.cpMaxBytes`** (new config, default 100 MiB, measured after excludes) are **blocked** unless **`--yes`**, with a `du`-style breakdown via **`dir-breakdown.ts`** (plus tests). Carry sizing, hashes, and the 50 MiB cap are exclude-aware so limits match what actually lands.
> 
> Docs and skills updated for cp/carry behavior and size limits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71d97651a2ac123c9047ed336e71b52513703b24. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Thanks Bugbot — triaged both:

- **File uploads skip size cap** (`cp.ts`) — **valid, fixed** in 96ab63d. `guardUploadSize` no longer early-returns for non-directories; single files are checked against `box.cpMaxBytes` too, with a file-appropriate message (can't split/exclude a file — only `--yes`). Verified live: a 120 MB file is now blocked without `--yes` and streams with it.
- **Carry cap wrong exclude paths** (`carry-resolve.ts`) — **false positive.** The relpath is `relative(dir, full)` where `dir` is the carry *root* (not the parent `d`), so it's the full multi-segment path, and `isPathExcluded('Application/cache', ['*/cache'])` matches correctly. The "one-segment from `relative(parent, child)`" reading doesn't match the code.

PR #93, opened 2026-06-22
fix(status): list tasks/services/ports in --inspect and cloud status

madarco:
## Problem

`agentbox status --inspect` (and plain `agentbox status` for **cloud boxes**, which always routes through the cloud inspect renderer since there's no host container to `docker exec` into) only printed a count line — e.g. `(0 svc, 4 tasks, 9 ports)` — and never listed the individual tasks. The per-task entries were already present in the persisted box-status snapshot; they just weren't rendered.

This regressed when the standalone `agentbox inspect` command was folded into `status --inspect` (commit `07f33f8`): only the one-line count survived, not the TASKS/SERVICES/PORTS sections that plain docker `status` still shows.

## Fix

- New `apps/cli/src/commands/_status-render.ts` with shared helpers: `fetchLive`, `renderLiveSections`, `renderPersistedSections` (extracted from `status.ts`; separate module avoids an import cycle with `inspect.ts`).
- `inspect.ts`: `renderText` (docker) and `renderCloudText` now append TASKS/SERVICES/PORTS — live from the in-box daemon when a docker box is running, else the persisted snapshot. Cloud renders the persisted snapshot.
- `status.ts`: reuses the shared helpers (behavior unchanged); `--inspect` help text fixed from "instead of service/task status" to "add ... to the service/task status".

## Test

- New `test/status-render.test.ts` (5 tests) asserting tasks are listed by name+state, ports roll up correctly, and the TASKS heading is omitted when there are none.
- Full apps/cli suite green (558 pass); `pnpm lint` + build clean.

---

> [!NOTE]
> **Low Risk**
> CLI display and refactor only; no auth, data, or runtime behavior changes beyond richer terminal output.
> 
> **Overview**
> Restores **TASKS / SERVICES / PORTS** detail in `agentbox status --inspect` and cloud status output, which had regressed to only a one-line persisted count after inspect was folded into status.
> 
> Shared logic moves to `_status-render.ts` (`fetchLive`, `renderLiveSections`, `renderPersistedSections`) so `status.ts` and `inspect.ts` stay in sync without an import cycle. **Docker inspect** appends live `agentbox-ctl` sections when the box is running (labeled as live), otherwise the persisted snapshot; **cloud inspect** always prints the persisted snapshot sections. Plain `status` behavior is unchanged aside from delegating rendering to the helpers; `--inspect` help now says it **adds** service/task status rather than replacing it.
> 
> Adds `status-render.test.ts` covering task names/states, port rollup, and omitting empty TASKS headings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c784bb70052bec1f9bf2d71efaa6298486857cb6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/_status-render.ts
+++ b/apps/cli/src/commands/_status-render.ts
@@ -1,0 +1,72 @@
+import {
+  renderPortsTable,
+  renderStatusTable,
+  renderTaskTable,
+  type BoxStatus,
+  type StatusReply,
+} from '@agentbox/ctl';
+import { execInBox } from '@agentbox/sandbox-docker';
+
+export async function fetchLive(state: string, container: string): Promise<StatusReply | null> {
+  // Only a running container is reachable via `docker exec` (macOS can see the
+  // socket file but can't connect to it). Paused/stopped — or a failed exec —
+  // falls back to the snapshot the relay persisted to disk.
+  if (state !== 'running') return null;
+  const proc = await execInBox(container, ['agentbox-ctl', 'status', '--json'], {
+    user: 'vscode',
+  });
+  if (proc.exitCode !== 0) return null;
+  try {
+    return JSON.parse(proc.stdout) as StatusReply;
+  } catch {
+    return null;
+  }
+}
+
+/** TASKS / SERVICES / PORTS blocks from a live `agentbox-ctl status` pull. */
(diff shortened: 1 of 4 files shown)


madarco:
bugbot run

madarco:
bugbot run

PR #101, opened 2026-06-23
feat: restore the running agent's session on box restart

madarco:
## What

When a box stops (or a cloud box idle-pauses and resumes), the agent's tmux session dies. On restart we re-run `agentbox.yaml` services but left the agent dead — the next attach started a **fresh** session, losing the conversation and any background/`-i` work.

This restores the running agent **resuming the same conversation** on `agentbox start`, on attaching to a down box, and on cloud idle-resume — so it looks like the box never stopped.

## How

- **Lazy session-id capture via the activity hooks.** Claude Code passes `session_id` on every hook payload, so the SessionStart/Stop hooks now run `agentbox-ctl claude-state … --capture-session`, recording it to a per-box pointer at `~/.local/state/agentbox/claude-session` — the box's own writable layer (not a shared volume), so it survives stop/start + cloud pause and tracks `/new` / `/branch`. Codex exposes no resumable id (and its hooks are unreliable), so the supervisor drops a `codex-active` marker instead.
- **On restart**, `restoreAgentSessions` (`apps/cli/src/agent-sessions.ts`) reads the pointers over `provider.exec` and relaunches: `claude --resume <id>` (exact) / `codex resume --last` (best-effort), idempotent (skips a live session). Skip-permissions applied from config.
- **Why exact-id, not `--continue`:** shared `~/.claude` / `~/.codex` volumes pool every box's sessions, so `--continue` / a file probe would resume the wrong session under parallel boxes and fire for agents that never ran in the box.
- A bare `agentbox <agent>` start stays fresh; OpenCode is skipped (no resume support yet).

## Verified end-to-end

| Provider | Claude (exact `--resume`) | Codex (`resume --last`) | Persistence |
| --- | --- | --- | --- |
| docker | ✅ prior conversation restored | ✅ | container stop/start |
| vercel | ✅ | — | survives **snapshot** |
| hetzner | ✅ | ✅ | survives **VPS** stop |

Unit test (`apps/cli/test/agent-sessions.test.ts`), ctl suite, typecheck, lint all green.

## Rollout note

The capture hooks are baked into the box image, so **docker needs a rebuild** and **cloud providers a re-`agentbox prepare`** (vercel + hetzner already re-prepared for the verification above; daytona/e2b would need it too). Until then restore no-ops — backward compatible.

https://claude.ai/code/session_017WPbmdQCPmqSr2iAxtHqXu

---

> [!NOTE]
> **Medium Risk**
> Touches box lifecycle, attach paths, and in-box hooks across docker and cloud; failures are logged and non-fatal, but wrong resume args or stale pointers could surprise users until hooks are deployed.
> 
> **Overview**
> Adds **session restore** so stop/start and cloud idle-resume do not drop the in-box agent conversation. Agent tmux dies with the container/VM; shared `~/.claude` / `~/.codex` volumes cannot identify which session belonged to which box, so restore uses **per-box pointer files** on the box writable layer (`~/.local/state/agentbox/`).
> 
> **In-box capture:** Claude hooks (`SessionStart` / `Stop`) call `agentbox-ctl claude-state --capture-session` to write `session_id` to `claude-session`. Codex gets a `codex-active` presence marker on first activity (no stable resume id). `StatusReporter` clears pointers when tmux goes from running → stopped so exited agents are not resurrected.
> 
> **Host restore:** New `agent-sessions.ts` exposes `agentResumeArgs` (`claude --resume <id>` / `codex resume --last`) and `restoreAgentSessions` (idempotent detached relaunch, best-effort, never fails `start`). Wired into `agentbox start`, cloud `unpause`/`dashboard` resume, and `claude attach` / `codex attach` / `cloudAgentAttach` when the box was down and the user passed no extra args. Bare `claude start` / `codex start` stay fresh. OpenCode unchanged.
> 
> Docs and a unit test for `agentResumeArgs` are included. **Existing boxes** need an image rebuild / cloud re-prepare for capture hooks; until then restore no-ops (backward compatible).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 507b2dd77f73679e0879f1f7f0d0b82be8ae44ec. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
Addressed in 5a6376d / 507b2dd / latest:
- Resync warning vs resume args: suppressed the seed-prompt injection when attach-resume injects resume flags (mirrors the teleport-resume path).
- Restore on cloud unpause: `restoreAgentSessions` now also fires on the cloud `provider.resume` path (`agentbox unpause` + dashboard); docker unpause is a cgroup thaw so tmux survives.
- Stale markers: the StatusReporter clears the pointer/marker on the running→stopped tmux edge, and Claude's SessionEnd hook now clears synchronously (no 15s race). Codex's marker stays best-effort (no reliable codex end signal).

PR #104, opened 2026-06-23
feat(e2b): configurable session timeout so the keepalive holds boxes open while working

madarco:
## What

E2B (like Vercel) auto-pauses a sandbox after its **session timeout**. The host relay's `cloud-keepalive` loop already pushes that timeout forward while the in-box agent is *working* — but E2B never had a configured/recorded timeout: `create.ts` only threaded `timeoutMs` into `providerOptions` for **vercel**, so an e2b box recorded `cloud.sessionTimeoutMs` as `undefined` (the keepalive fell back to a 45-min *assumption*) and the lifetime wasn't tunable. This adds the parity needed so E2B's "keep alive while the agent works" behaves exactly like Vercel.

## How

- **`box.e2bTimeoutMs`** config key (mirrors `box.vercelTimeoutMs`, default 45 min).
- A shared **`cloudSizingProviderOptions(providerName, cfg)`** helper (`apps/cli/src/lib/cloud-sizing.ts`), used by `agentbox create` **and** the agent-create commands (`claude`/`codex`/`opencode`), so an e2b box records its real session timeout (which seeds the keepalive precisely) regardless of how it was made. Bonus: this also threads **vercel's** sizing through the agent-create path, which it wasn't before.

No keepalive-loop change was needed — it's already generic and E2B implements `renewTimeout` → `Sandbox.setTimeout`.

## Verified live (E2B)

Created an e2b box with a 2-min timeout, drove a `working` agent state:
- relay logged `cloud-keepalive: renewed box <id> (e2b) +5m` **3×** while working;
- the sandbox stayed **alive past its 2-min create timeout** (would've died without it);
- once **idle**, renewals **stopped** (0 in 3 min) — the timer is only held open while working.

`cloud.sessionTimeoutMs` was correctly recorded as `120000` on the box.

## Notes

- Pure host-side + config change — **no box image rebuild or `prepare` needed**.
- Caveat (documented): E2B **Hobby tier ~1 h platform cap** bounds total session regardless of activity; the interactive attach PTY still caps at 55 min (separate from box lifetime — reattach to continue).
- Unit test `apps/cli/test/cloud-sizing.test.ts`; docs updated (`e2b_backlog`, `cloud-providers` §3c, web config reference).

https://claude.ai/code/session_017WPbmdQCPmqSr2iAxtHqXu

---

> [!NOTE]
> **Low Risk**
> Host-side config and CLI `providerOptions` threading only; keepalive behavior is unchanged and scoped to cloud create paths. Mis-set timeouts could shorten or lengthen E2B box lifetime but do not touch auth or data handling.
> 
> **Overview**
> Adds **`box.e2bTimeoutMs`** (default 45 min, E2B-only) so new `--provider e2b` boxes get a tunable create-time session timeout, recorded as **`cloud.sessionTimeoutMs`** for the existing host **`cloud-keepalive`** loop instead of relying on an assumed deadline when that field was missing.
> 
> Introduces **`cloudSizingProviderOptions`** in `apps/cli/src/lib/cloud-sizing.ts` and wires it through **`agentbox create`** plus cloud **`claude` / `codex` / `opencode`** creates. **`create`** replaces inline Vercel-only `providerOptions` with this helper (Vercel vcpus/timeout/network policy unchanged; E2B gains `timeoutMs`). Agent-create paths now pass the same sizing overrides as **`create`**, including Vercel options that were not threaded there before.
> 
> Config types/defaults/schema, unit tests, and docs (`e2b_backlog`, `cloud-providers`, web config reference) are updated to describe E2B box lifetime vs the 55 min attach PTY cap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f29810a8218573d5ee5a7f17e0012055baeeeb9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/claude.ts
+++ b/apps/cli/src/commands/claude.ts
@@ -37,6 +37,7 @@ import {
 import { Command } from 'commander';
 import { resolveClaudeAuth, type ResolvedClaudeAuth } from '../auth.js';
 import { reattachRef, resolveBoxOrExit, resolveBoxOrShift } from '../box-ref.js';
+import { cloudSizingProviderOptions } from '../lib/cloud-sizing.js';
 import { assertAgentCredsAvailable, MissingAgentCredsError } from '../lib/queue/assert-creds.js';
 import { buildPromptArgs } from '../lib/queue/build-prompt-args.js';
 import { maybeResyncWorkspace } from '../lib/resync-start.js';
@@ -782,6 +783,9 @@ export const claudeCommand = new Command('claude')
           useBranch,
           resyncOnStart: opts.resync,
           projectRoot,
+          // Per-provider session-lifetime (e2b/vercel timeout) so the keepalive
+          // seeds correctly; mirrors `agentbox create`.
+          providerOptions: cloudSizingProviderOptions(provider.name, cfg.effective),
         },
         binary: 'claude',
         hasSeedPrompt:
--- a/apps/cli/src/commands/codex.ts
+++ b/apps/cli/src/commands/codex.ts
@@ -48,6 +48,7 @@ import { buildPromptArgs } from '../lib/queue/build-prompt-args.js';
 import { maybeResyncWorkspace } from '../lib/resync-start.js';
 import { buildResyncWarning } from '../lib/resync-warning.js';
 import { agentResumeArgs } from '../agent-sessions.js';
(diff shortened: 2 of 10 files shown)


madarco:
bugbot run

PR #112, opened 2026-06-25
Support git-lfs repos in the box

HageMaster3108:
Box creation failed for any repo whose host ~/.gitconfig configures the git-lfs filter (filter.lfs.process = git-lfs filter-process, set by `git lfs install`). The box image installs git but not git-lfs, and the host ~/.gitconfig is bind-mounted RO into the box, so the in-container `git worktree add` (seedWorkspace) spawned the filter on checkout and died:

    git-lfs filter-process: 1: git-lfs: not found
    fatal: the remote end hung up unexpectedly

Two changes:

- Dockerfile.box: install git-lfs and `git lfs install --system` so the filter binary resolves on PATH. --system writes /etc/gitconfig (the host-config bind only covers ~/.gitconfig), so repos that declare LFS purely via .gitattributes resolve too.

- in-box-git.ts: before the in-box `git worktree add` checkout (and the checkpoint-restore `reset --hard`), best-effort `git lfs fetch` the base ref host-side. The host has the credentials and .git/lfs/objects is part of the bind-mounted .git shared with the box, so the in-box smudge then resolves every object from cache with no box-side network or credentials. A non-LFS repo is a silent no-op; any failure is logged and swallowed so create never aborts.

### DIFF
--- a/packages/sandbox-docker/Dockerfile.box
+++ b/packages/sandbox-docker/Dockerfile.box
@@ -68,6 +68,7 @@ RUN apt-get update \
         python3-venv \
         build-essential \
         git \
+        git-lfs \
         tmux \
         vim \
         libcap2-bin \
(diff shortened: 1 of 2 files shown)


madarco:
thanks git-lfs it was high on my list!

madarco (approved):
Validated on docker: git-lfs installed in box image + filter registered system-wide; host-side `git lfs fetch origin` prefetch verified end-to-end against ssh and https LFS remotes (credential-less box resolved objects from the shared bind-mounted .git/lfs cache, oids matched).

PR #115, opened 2026-06-25
Cloud: fall back to xterm-256color when the box lacks the host TERM

madarco:
Ports the docker TERM-fallback fix (#113) to the **cloud providers** — daytona, hetzner, vercel, e2b. Targets `nightly` (where #113 landed).

## Problem

The shared `renderInnerCommand` (`packages/sandbox-cloud/src/cloud-provider.ts`) builds the in-box tmux attach command for **all four** cloud providers, and had no TERM guard:

- **hetzner / daytona** attach over `ssh -t`, which forwards the host TERM as-is. When the box's Ubuntu terminfo lacks that TERM (e.g. `xterm-ghostty`, added to ncurses after 24.04 shipped), `tmux attach` exits immediately with *"missing or unsuitable terminal"* — the same flash-and-quit #113 fixed for docker. **This was the live bug.**
- **vercel / e2b** hardcoded `TERM=xterm-256color`, so they never hit the bug but also dropped host-terminal fidelity.

## Fix

- Export `TERM_FALLBACK_SNIPPET` from `sandbox-docker` (single source of truth) and prepend it in `renderInnerCommand` before the tmux commands — **one change covers all four providers**. The guard is a no-op when the box already knows `$TERM`, and safely downgrades to `xterm-256color` when it doesn't (or when `infocmp` is absent), so it never regresses the providers that hardcoded TERM.
- Forward the host TERM on **vercel** (`build-attach` envPrelude) and **e2b** (`build-attach` → `AGENTBOX_HOST_TERM` → `attach-helper` PTY env) so a box that carries the host terminfo renders at full fidelity, matching docker. hetzner/daytona already forward via `ssh -t`.

## Notes

- **Host-side only — no base-image rebake needed.** The guard relies on `infocmp`/ncurses, already present in every base image; and the fallback is safe even if it weren't.
- Build + lint + typecheck green. Added a cloud guard test (`render-inner-command.test.ts`) and updated the vercel/e2b `build-attach` assertions. docker 297 / cloud 81 / vercel 82 / e2b 16 tests pass.

## Verification status

Unit-tested + statically verified. Live cloud e2e (create a hetzner/vercel box, attach with `TERM=xterm-ghostty`, confirm it renders instead of flash-quitting) is cheap here since no rebake is required — can run on request.

https://claude.ai/code/session_01PTY4KwAeZdAVvgSWxjpYfs

---

> [!NOTE]
> **Low Risk**
> Changes only interactive attach command strings and TERM env wiring; behavior is defensive (fallback to xterm-256color) with new unit tests and no security or data-path impact.
> 
> **Overview**
> Fixes **flash-and-quit** cloud attaches when the host forwards a `TERM` the box’s Ubuntu terminfo does not know (e.g. `xterm-ghostty` over Hetzner/Daytona `ssh -t`).
> 
> **Shared guard:** `TERM_FALLBACK_SNIPPET` is **exported** from `@agentbox/sandbox-docker` and **prepended** to `renderInnerCommand` for every tmux path (interactive attach and detached pre-start). It runs `infocmp` and downgrades to `xterm-256color` when needed; **logs** and **noTmux** are unchanged.
> 
> **Host TERM forwarding:** Vercel’s attach prelude and E2B (`AGENTBOX_HOST_TERM` → in-box PTY) now set **`TERM` from the host** instead of always `xterm-256color`, with the same guard handling unknown terms.
> 
> New unit coverage in `render-inner-command.test.ts`; Vercel/E2B attach tests updated for the new prelude/env.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 051af1babcb3af6155c25a9fec2d86d62dbde37e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sandbox-cloud/src/cloud-provider.ts
+++ b/packages/sandbox-cloud/src/cloud-provider.ts
@@ -48,6 +48,7 @@ import {
   portlessGetUrl,
   portlessUnalias,
   registerBoxWithRelay,
+  TERM_FALLBACK_SNIPPET,
 } from '@agentbox/sandbox-docker';
 import {
   ensureAgentVolumesForCloud,
@@ -1425,6 +1426,15 @@ function makeCloudCheckpoint(backend: CloudBackend): ProviderCheckpoint {
  * starts the session in the box's workspace dir so claude/codex/opencode
  * see /workspace as their cwd (otherwise tmux inherits the SSH login
  * shell's $HOME and the agents prompt for workspace-trust).
+ *
+ * The command is prefixed with the shared {@link TERM_FALLBACK_SNIPPET} (same
+ * guard the docker provider applies): hetzner/daytona forward the host TERM
+ * over `ssh -t`, and vercel/e2b forward it via their attach builders, so a
+ * host TERM the box's terminfo doesn't carry (e.g. xterm-ghostty on Ubuntu)
+ * would otherwise make `tmux attach` exit immediately ("missing or unsuitable
+ * terminal"). The guard downgrades the unknown case to xterm-256color before
+ * tmux runs, and is a no-op (`infocmp` missing → also downgrades) when the box
+ * lacks `infocmp`, so it never regresses the providers that hardcoded TERM.
  */
 export function renderInnerCommand(kind: AttachKind, opts?: BuildAttachOptions): string {
   const sessionName = opts?.sessionName ?? defaultSessionName(kind);
@@ -1448,8 +1458,8 @@ export function renderInnerCommand(kind: AttachKind, opts?: BuildAttachOptions):
   // `detached`: create + configure the session but don't attach. Used to
   // pre-start a session with its full launch command before a new-tab attach
   // re-invokes `agentbox <agent> attach` (which carries no launch args).
-  if (opts?.detached) return lines.join('; ');
-  return [...lines, `exec tmux attach -t ${sessionQ}`].join('; ');
+  if (opts?.detached) return TERM_FALLBACK_SNIPPET + lines.join('; ');
+  return TERM_FALLBACK_SNIPPET + [...lines, `exec tmux attach -t ${sessionQ}`].join('; ');
 }
 
 function defaultSessionName(kind: AttachKind): string {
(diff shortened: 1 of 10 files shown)


madarco:
## Live cloud verification ✅

Tested on every runnable cloud provider against a real box (no rebake needed — host-side only). For each: confirmed the box's terminfo **lacks** `xterm-ghostty` (so the bug triggers), has `infocmp` + `xterm-256color` (guard + fallback target), then drove a real PTY (`script`) attach with `TERM=xterm-ghostty` forwarded.

| Provider | pre-fix (raw `xterm-ghostty`) | with guard | verdict |
|---|---|---|---|
| **hetzner** | `missing or unsuitable terminal: xterm-ghostty` (flash-quit) | resolves `TERM=xterm-256color`, tmux attaches | **PASS** |
| **e2b** | `missing or unsuitable terminal: xterm-ghostty` | resolves `xterm-256color`, attaches | **PASS** |
| **vercel** | `missing or unsuitable terminal: xterm-ghostty` | resolves `xterm-256color`, attaches | **PASS** |
| **daytona** | — (no creds in this env) | shares the identical `renderInnerCommand` guard + `ssh -t` transport as hetzner | covered by shared code path |

Also confirmed via the real CLI attach path (drive harness, `TERM=xterm-256color`) that the **normal** case still attaches cleanly on all three (`client_term=xterm-256color`, session attached) — no regression.

Note: the drive PTY harness pins node-pty `name: 'xterm-256color'`, so it can't itself carry an exotic TERM; the `script`-based in-box probe was used to exercise the guard with a real `xterm-ghostty`.

https://claude.ai/code/session_01PTY4KwAeZdAVvgSWxjpYfs

madarco:
Addressed the Bugbot finding (unquoted host TERM in the vercel `bash -lc` prelude) in 167c5b8: added `hostTermForCloud()` which only forwards a well-formed terminfo name (`[A-Za-z0-9][A-Za-z0-9._+-]*`) and falls back to xterm-256color otherwise. Both vercel and e2b now route through it. Added sanitization unit tests (rejects spaces / `$(...)` / backticks / `;`).

PR #119, opened 2026-06-26
feat(codex): expose box system-prompt to Codex via ~/.codex/AGENTS.override.md

madarco:
## What

The box \"system prompt\" baked at \`/etc/claude-code/CLAUDE.md\` (sandbox facts: DinD, per-box git worktree, push/PR/cp via the host relay, identity in \`/etc/agentbox/box.env\`) previously only benefited **Claude**. **Codex got none of it.** This wires the same facts into Codex's global-instructions slot, **composed** with the user's own Codex global, across **all providers**.

## How

Codex loads a global personal-instructions file from \`CODEX_HOME\`, first-match of \`~/.codex/AGENTS.override.md\` then \`~/.codex/AGENTS.md\` (verified in \`openai/codex\` source — pure whole-file concat, **no \`@import\`**). At create time we regenerate \`~/.codex/AGENTS.override.md\` = sentinel line + box facts (read fresh) + the user's own \`AGENTS.md\`/authored-override folded in beneath.

- A line-1 sentinel makes it idempotent and prevents folding our own output back in / destroying authored content.
- The host \`~/.codex\` is re-synced before each seed (docker), so the user source is restored every create and the fold stays stable.
- One shared generator \`buildCodexAgentsOverrideScript()\` drives both paths:
  - **docker** — \`seedCodexAgentsOverride()\` seeds the codex-config volume, called after \`seedCodexHooks()\` (post host rsync).
  - **cloud** (daytona/hetzner/vercel/e2b) — \`ensureCodexAgentsOverride()\` runs the same script in-box via \`backend.exec\`, wired into \`cloud-provider.ts\`.

Why a generated file (not a symlink or \`@import\`): Codex's loader is first-match-wins, so a symlink would **shadow** a user's \`AGENTS.md\`, and \`@AGENTS.md\` would reach the model as dead text. Auto-concatenation is the faithful \"compose without hand-maintaining a copy.\"

## Verification

- Real docker box: override is \`vscode:vscode 644\`, sentinel line 1, body byte-matches \`/etc/claude-code/CLAUDE.md\`.
- **\`codex debug prompt-input\` shows the box facts in Codex's model-visible prompt** (\"You are running inside an AgentBox sandbox…\").
- Compose / authored-override / facts-only / facts-missing-no-op cases all verified; 385 package tests + full lint green.

Known boundary (documented in the docstring): an override hand-authored *inside* the box's shared volume with no host counterpart isn't a durable surface — the supported source is the host \`~/.codex\`, re-synced each create.

https://claude.ai/code/session_01PTY4KwAeZdAVvgSWxjpYfs

---

> [!NOTE]
> **Low Risk**
> Create-time, best-effort file generation in agent config only; no auth, relay, or credential path changes.
> 
> **Overview**
> **Codex now receives the same sandbox “box facts” as Claude** by folding `/etc/claude-code/CLAUDE.md` into `~/.codex/AGENTS.override.md` at box create time on **Docker and cloud** providers.
> 
> A shared shell generator (`buildCodexAgentsOverrideScript`) writes a sentinel line, the box-facts file, then the user’s own `AGENTS.md` or hand-authored override beneath—because Codex uses first-match-wins for global instructions, not concatenation. **Docker** seeds via `seedCodexAgentsOverride` on the codex-config volume after the host `~/.codex` rsync; **cloud** runs the same script in-box through `ensureCodexAgentsOverride` (with `chown` to `vscode`). Failures are best-effort and do not block create.
> 
> Docs and `Dockerfile.box` comments are updated to describe Codex loading this path versus Claude’s env-only discovery today.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3ff4616c5b432822cae70c4b453fa54f808fc31. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
Thanks Bugbot. Triage:

**1. "Cloud seed ignores exec exit" — fixed** in f3ff4616c. Cloud `backend.exec` returns a non-zero `CloudExecResult.exitCode` on script failure rather than throwing, so the helper now reads `res.exitCode` and logs the failure instead of unconditionally reporting success. Still best-effort (never fails create).

**2. "Cloud reads stale box facts" — by design, not changing.** At create time the box-facts file (`/etc/claude-code/CLAUDE.md`) exists only inside the box's own filesystem — it's a baked artifact and the host has no copy — so reading the live sandbox is the only source available to the cloud path, and it's exactly what every other in-box tool sees (as fresh as the snapshot). Docker reads from `ensureRef` only because it seeds the codex-config volume in a throwaway container *before* the box exists. Re-uploading the facts file from the repo on every cloud create to chase the docker-vs-snapshot freshness asymmetry isn't worth it for content that rarely changes; this matches the existing docker-rebuilds-vs-cloud-snapshots model documented in docs/cloud-create-flow.md.

bugbot run

madarco:
bugbot run

madarco:
**"Cloud logs seed without file" — fixed** in f64253378. The generator now prints a stdout marker (`agentbox:codex-override-written`) only after it actually writes the override; the early `exit 0` (box-facts file absent) skips it. Both docker and cloud key their "seeded" log off the marker, and cloud logs `skipped: box-facts file absent` on a 0-exit no-op. In practice the file is always present (installed at prepare time on every provider), so this is defensive log accuracy.

The remaining "stale box facts" item is intentional/by-design as explained above (the box's own fs is the only source available to the cloud seed at create time).

bugbot run

PR #129, opened 2026-06-29
feat(hetzner): self-heal the per-box firewall on a host egress-IP change

madarco:
## Problem

A Hetzner box's per-box Cloud Firewall locks SSH to the host's egress IP **at create time** and is never re-synced. If the host IP changes (laptop moves networks), the firewall silently drops SSH from the new IP and every comms op — `exec`, `attach`, port-forward, the cloud poller, `recover` — fails with an **opaque** `ssh ControlMaster failed … connect to host <ip> port 22: Operation timed out`. The fix exists (`agentbox hetzner firewall sync`), but nothing hints at it and `recover` doesn't auto-heal.

## Fixes

Both gated to the **connection-failure path** (the happy path never runs the egress-detect curls), and the firewall is re-synced **only when the IP actually changed**.

1. **Hint (read-only)** — wrap `tunnels.open` in `ensureTunnel`, the single choke point all of exec/scp/forward/poller/**attach** funnel through. On a real mismatch it replaces the opaque timeout with *"firewall allows X but your egress is now Y — run `firewall sync` / `recover`"*. Safe on a checkpoint drop (box merely stopped, IP unchanged → no hint).

2. **Auto-sync, scoped to connection ESTABLISHMENT only** — new optional `repairReachability` on `CloudBackend`/`Provider` (Hetzner-only) re-syncs the firewall to the current egress, returning `{changed:false}` when it didn't change. A `withFirewallRepair` CLI helper retries the attempt once **iff** something changed, wired at exactly two establish sites:
   - `recover` → `provider.reconnect`
   - the **initial** attach connect → `_cloud-attach` `buildAttach`

   …and **deliberately not** the mid-session reconnect closure — a `checkpoint --set-default` stops the box and drops the PTY, which must never be mistaken for an IP change. `--no-firewall-sync` opts out on `recover` (shared/untrusted egress).

A short-TTL egress cache avoids probe storms across retries / `recover --all`. A `0.0.0.0/0` firewall (explicit dynamic-IP opt-in) is never hinted or synced.

## Verification

- **Live (Hetzner box)**: locked the firewall to a bogus `1.2.3.4/32` →
  - `agentbox shell … -- echo` fails with the **hint** and does NOT auto-repair;
  - `agentbox recover …` auto-syncs back to the real egress + reconnects ("firewall updated: SSH now allowed from … (was 1.2.3.4/32)");
  - `--no-firewall-sync` leaves it locked (WARN).
- **Unit**: `firewallNeedsSync` (match→false, change→true, `0.0.0.0/0`→false, absent→true) and the egress TTL cache (probe once within window, re-probe after).
- `pnpm typecheck && lint && test` green.

https://claude.ai/code/session_01Ja5HgEjwyER5BhhFCpPUup

---

> [!NOTE]
> **Medium Risk**
> Changes SSH reachability and firewall rules on connection failures; scoped to Hetzner establish paths with one retry and opt-out, but mis-sync could briefly widen or mis-target firewall rules if egress detection is wrong.
> 
> **Overview**
> When a laptop’s public IP changes, Hetzner boxes keep SSH locked to the old egress CIDR, so attach/exec/recover fail with opaque SSH timeouts.
> 
> This PR adds **connection-establishment-only** self-heal: optional `repairReachability` on `CloudBackend`/`Provider` (Hetzner implements it), plus CLI helper `withFirewallRepair` that retries once after a successful firewall sync.
> 
> **Auto-sync** runs on `agentbox recover` (`provider.reconnect`) and on **initial** Hetzner cloud attach (up-front `exec true` before resume probe / `buildAttach`). Mid-session attach **reconnect** deliberately does **not** sync—the box stopping is not an IP change. `recover` gains `--no-firewall-sync` to opt out.
> 
> On tunnel open failure, Hetzner **`ensureTunnel`** compares firewall SSH source vs fresh egress and appends a hint to run `hetzner firewall sync` or `recover`. Pure `firewallNeedsSync` (skips `0.0.0.0/0`) is unit-tested.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec544984caf21981ee87d3740837fab17a549499. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/_cloud-attach.ts
+++ b/apps/cli/src/commands/_cloud-attach.ts
@@ -186,6 +186,24 @@ export async function cloudAgentAttach(args: CloudAgentAttachArgs): Promise<void
     box = await provider.start(box);
     s.stop('box running');
   }
+  // Hetzner only: open the SSH tunnel UP FRONT, self-healing a stale firewall (a
+  // host egress-IP change locks the per-box firewall) BEFORE any of the later
+  // establish touches — the resume probe, the detached pre-start, buildAttach.
+  // Whichever of those connected first would otherwise be an unguarded
+  // establish: a firewall block there aborts the attach (or silently drops the
+  // resumed session, since the resume probe swallows exec errors). Doing it once
+  // here covers them all. Repairs ONLY on an actual connect failure; otherwise a
+  // `true` over the already-open master is a cheap no-op. This is an ESTABLISH
+  // path — distinct from the mid-session `reconnect` closure below, which must
+  // NOT touch the firewall (a checkpoint/pause drop isn't an IP change).
+  if (box.provider === 'hetzner') {
+    await withFirewallRepair(
+      provider,
+      box,
+      { enabled: true, onLog: (line) => log.success(line) },
+      () => provider.exec(box, ['true']),
+    );
+  }
   // Attaching to a box that just came back up (a stop / cloud idle-timeout
   // resume): if the user passed no args of their own and the box has a resumable
   // claude/codex session, launch resuming it (claude --resume <id> / codex resume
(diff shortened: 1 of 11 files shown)


madarco:
bugbot run

madarco:
bugbot run

PR #135, opened 2026-07-01
feat: add box.claudeInstall to install Claude via npm at bake time

madarco:
## Why

Every provider bakes Claude Code with Anthropic's **native installer** (`curl claude.ai/install.sh`). Its CDN intermittently **403s cloud-datacenter egress IPs** (Hetzner especially) under load. The bake retries 3× then hard-fails (`exit 71`) rather than ship a Claude-less base — which leaves users on affected IPs unable to bake at all.

This adds an **opt-in escape hatch**: `box.claudeInstall` (enum `native` default | `npm`) switches the bake to install `@anthropic-ai/claude-code` from npm instead. Native stays the default (recommended path, native-only features).

## How

- **Path reconciliation:** npm lands `claude` at a different path (`/usr/bin/claude`) than the native `/home/vscode/.local/bin/claude` that everything downstream hardcodes (attach command, PATH shim, host-side `installMethod=native` coercion). The npm branch **symlinks** it into that path so the box stays indistinguishable from a native install.
- **Bake-time only** — read by `agentbox prepare`, never `create`. The mode is folded into each provider's base fingerprint (`claudeInstallFingerprint`, `@agentbox/sandbox-core`) so switching re-bakes; `native` leaves the hash byte-identical (no spurious re-bake for existing users).

Wired across **all bake providers**:
- **hetzner / vercel / e2b** — install script branches on `AGENTBOX_CLAUDE_INSTALL`.
- **docker** — Dockerfile `ARG`/`--build-arg`; `ensureImage` folds the mode into its create-time fingerprint so the lazy rebuild doesn't clobber an npm image.
- **daytona** (SDK has no build-arg) — builds from a sibling temp Dockerfile with the `ARG` default flipped (relative COPY paths still resolve).

CLI: `agentbox prepare --claude-install <native|npm>` (wins over the config key).

## Verification

- `pnpm build` (15/15), `pnpm lint` (clean), full `pnpm test` — all touched packages green. Added schema-drift fixtures for the new enum.
- CLI round-trip: default `native`; set/get `npm`; invalid values rejected by both the config parser and the `--claude-install` flag.

### Live-verify gate (needs a real bake, not covered by unit tests)
Whether Claude Code's own startup integrity check stays quiet with an npm binary symlinked at the native path. Fallback if it complains: set the box's `.claude.json` `installMethod` in npm mode only. To test: `agentbox config set box.claudeInstall npm && agentbox prepare --provider hetzner --force`, then run a real `claude -p` turn in a box.

https://claude.ai/code/session_01TGoDQvktdXD5rqJuJm9Uyk

---

> [!NOTE]
> **Medium Risk**
> Touches every provider’s base-image bake and Docker lazy rebuild fingerprints; wrong wiring could ship Claude-less bases or clobber npm images with native pulls, but scope is opt-in and defaults stay native.
> 
> **Overview**
> Adds **`box.claudeInstall`** (`native` | `npm`, default `native`) and **`agentbox prepare --claude-install`** so bake-time Claude Code installation can use **`@anthropic-ai/claude-code`** when Anthropic’s native installer CDN returns **403** on datacenter egress (e.g. Hetzner). The setting is **bake-only** (`prepare`, not `create`); CLI overrides config.
> 
> **`claudeInstallFingerprint`** folds `npm` into base snapshot/image fingerprints so switching modes triggers a re-bake without changing fingerprints for existing native bases.
> 
> **All bake paths** honor the mode: shell bakes get **`AGENTBOX_CLAUDE_INSTALL`**; Docker uses **`Dockerfile.box` `ARG`** + build-args and disables registry pull for npm; Daytona builds from a **temp Dockerfile** with the ARG default flipped; **`ensureImage`** reads config so lazy Docker rebuilds don’t overwrite an npm image. The npm path **symlinks** `claude` to **`~/.local/bin/claude`** to match native layout.
> 
> Docs and JSON schema cover the new key; schema-drift tests validate the enum.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2acaecc3b50a55c37b9beca87a45adab2b405d1c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
## Two provider-prepare fixes added (smoke-validated)

**1. `fix(hetzner): forward claudeInstall` (9af4f9ed6)** — addresses the Cursor Bugbot High-severity finding above. `prepareHetznerProvider` wasn't forwarding `req.claudeInstall`, so the bake always ran the native `curl install.sh` → Cloudflare 403 → `exit 71`. Now forwarded (matches `prepareVercelProvider`). **Smoke:** `agentbox prepare --provider hetzner --force` now runs the **npm** install and **completes the bake** (writes `box.imageHetzner` + `hetzner-prepared.json` 0.21.0) — previously aborted on the 403.

**2. `fix(e2b): bake host static agent config via shared sync-layer stager` (ab980b0b4)** — e2b's template bake never staged host static config, so e2b boxes lacked `~/.claude` settings/plugins + `~/.codex/config.toml`. Hoisted the producer→box-dir mapping into `sandbox-core/src/sync/host-stage.ts` as `stageAllAgentStatic()` (single source of truth for all cloud prepare paths) and consumed it in e2b. **Smoke (real e2b box):** `~/.codex/config.toml` 1466B (sanitized, `/workspace` trusted), `~/.claude/settings.json` 2013B (host hooks stripped), 6 plugins — full parity with docker/vercel.

**Gate:** build + typecheck + lint green; core/sandbox-core/docker/cloud/relay/cli tests unaffected.

### Follow-ups (separate, pre-existing — not in this PR)
- **Hetzner box boot:** after the (now-succeeding) bake, `agentbox create --provider hetzner` fails with `agentbox-ctl: unknown command 'bootstrap'` + "base image out of date" — the baked snapshot carries a **stale in-box ctl** (base-staleness / incremental-base skew), orthogonal to the 403 fix. Needs a separate look at hetzner prepare refreshing the ctl binary.
- **Daytona static seeding** still partial (settings stub, no codex config.toml) despite prepare appending the right `addLocalFile`+`runCommands` steps to `Image.fromDockerfile(...)` — the Daytona SDK build isn't landing them. Deferred pending a verbose-build reproduction.

madarco:
## Fixed the "base image out of date" false-positive (cf602896c)

Root-caused the hetzner "stale-ctl / `agentbox-ctl: unknown command bootstrap`" report. It was **two things, neither a ctl staging bug**:

1. **Real bug (this commit):** the claudeInstall feature made `prepare` fold `box.claudeInstall` into the stored base fingerprint (`claudeInstallFingerprint`), but the create-time freshness recompute (`current<Provider>BaseFingerprintLive`) returned the **raw** contextSha. So on any `claudeInstall: npm` machine, *every* freshly-baked cloud base perpetually read **"base image out of date"** — for **all four** cloud providers. Threaded the resolved mode through `evaluateBaseFreshness → currentCloudBaseFingerprintLive → Provider.baseFingerprint(claudeInstall)` and fold it in each provider's live recompute, mirroring `prepare`. `native` is identity (no change for native users); `npm` now matches its baked base.

2. **Testing artifact (not a product bug):** the actual `bootstrap`-unknown boot failure came from preparing in one project but creating with `--workspace <another>` that held a stale per-project image pin (old 0.20.1 snapshot). A same-project fresh bake boots fine.

**Validated:** with the fix, a fresh `claudeInstall: npm` hetzner bake → `create` shows **no "out of date" warning** and boots cleanly (`bootstrap` present, codex `config.toml` 1466B, claude settings 2013B, 6 plugins). Build/typecheck/lint + all provider + cli tests green.

madarco:
## Fixed the last provider — daytona static seeding (a996d1dd0)

**Root cause (Daytona SDK):** daytona built its snapshot with `Image.fromDockerfile(...).addLocalFile(absTmpTarball, remoteTar).runCommands(tar…)`. `addLocalFile` emits `COPY <absolute-tmp-path>` but archives the context entry under the **stripped-relative** name (node-tar drops the leading `/`). The Daytona builder resolves `COPY` sources relative to the reconstructed context root, so the absolute source never matched the relative archive entry and the whole seed layer **silently dropped** — daytona boxes shipped a 37-byte stub `settings.json` and no `~/.codex/config.toml`. (`fromDockerfile`/`dockerfileCommands` strip the prefix to a *relative* archive path, which is why Dockerfile.box's own COPYs work.)

**Fix:** stage the seed files (tarballs + daytona CLAUDE.md) into a temp build-context dir under **relative** names and append them via `dockerfileCommands(cmds, seedContextDir)` with relative COPY sources — the same relative-archive model e2b uses. Also:
- switched daytona onto the shared sync-layer `stageAllAgentStatic` (dropped its local duplicate);
- folded a `DAYTONA_SEED_SCHEMA` bump into `computeDaytonaContextFingerprint` so existing snapshots re-bake once (context files unchanged, only the build logic);
- extracted `buildDaytonaSeedCommands` + a regression test asserting every COPY source is relative (guards against regressing to `addLocalFile`).

**Validated on a real daytona box** (created in the same project it was prepared in — a cross-project stale pin will still boot an old snapshot, same as hetzner): `~/.claude/settings.json` 2013B (host hooks stripped) + 6 plugins, `~/.codex/config.toml` 1466B (sanitized, `/workspace` trusted), opencode + `~/.agents` present — parity with docker/vercel/e2b. Gate: build/typecheck/lint + daytona tests (incl. new) green.

All four cloud providers now bake host static config correctly.

PR #163, opened 2026-07-07
fix(portless): adopt a running proxy on non-interactive box creates

madarco:
## Problem

On a fresh Mac, starting the **first** box from the tray app does **not** get a Portless `https://<box>.localhost` URL — even though the Portless proxy is running on :443. Running `agentbox claude` once from a terminal (which shows the interactive "Use Portless?" prompt, answered Yes) fixes it: from then on every box, app-created included, gets Portless.

## Root cause

The tray app creates boxes via the Control Hub (`POST /api/v1/boxes`), which enqueues a **background queue job**. That path never runs the interactive Portless opt-in — the queue worker resolved Portless from config only:

```ts
// _run-queued-job.ts (before)
portless: opts.portless ?? cfg.effective.portless.enabled,   // undefined on a fresh machine → no Portless
```

`portless.enabled` is unset until the user opts in once from a real terminal (the only path that prompts + persists `portless.enabled: true`). So the first app-created box got no `.localhost` alias despite a live proxy. The foreground non-interactive CLI path (`maybePromptPortless`) had the same blind spot: `if (!process.stdin.isTTY || args.yes) return false`.

## Fix

Add a shared `resolvePortlessNonInteractive()` helper: when Portless is undecided and the engine is Docker Desktop, **adopt an already-running proxy** (persisting the choice so later runs skip re-detection) — but never install or start one unasked, and stay off for OrbStack (which has `.orb.local`). Wired into both non-interactive entry points:

- `maybePromptPortless` — the `--yes`/no-TTY branch now resolves through it instead of returning `false`.
- `_run-queued-job.ts` — the queue worker (the tray/hub create path) resolves through it before `createBox`.

## Verification

- **Live e2e** in the exact repro environment (Docker Desktop + Portless on :443, `portless.enabled` unset to simulate a fresh Mac): non-interactive `create -y` registered `https://portless-smoke.localhost` + `https://vnc-portless-smoke.localhost` and re-persisted `portless.enabled: true`. Before the fix this path returned `false` and registered no alias.
- **9 new unit tests** (`apps/cli/test/portless-prompt.test.ts`) covering both entry points: adopt-when-running, skip-when-down, honor explicit config/flag, OrbStack stays off.
- `typecheck` + `eslint` clean.

No behavior change for users who already opted in (config short-circuits), for `--no-portless`, or for OrbStack.

---

> [!NOTE]
> **Low Risk**
> Behavior-only change for undecided Portless on non-interactive paths; explicit config, flags, and OrbStack are unchanged, and no proxy is started without user consent.
> 
> **Overview**
> Fixes tray-app / hub **first box** missing `https://<box>.localhost` when Portless is already on `:443` but `portless.enabled` was never set (only the interactive CLI prompt persisted it).
> 
> Adds **`resolvePortlessNonInteractive`**: if Portless is undecided, honor explicit flags/config; skip OrbStack; on Docker Desktop **enable Portless when a proxy is already running** and best-effort persist `portless.enabled: true` — without installing or starting Portless. Wires it into **`maybePromptPortless`** (no TTY / `--yes`) and the **queue worker** (`_run-queued-job`) before `createBox`.
> 
> New unit tests cover adopt-when-running, no-proxy, explicit config, and OrbStack.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80dee039fb9f482e376cc54f0d1a27761d1ba6cd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/_run-queued-job.ts
+++ b/apps/cli/src/commands/_run-queued-job.ts
@@ -24,6 +24,7 @@ import {
 import {
   createBox,
   DEFAULT_BOX_IMAGE,
+  detectEngine,
   ensureImage,
   hostBackupHasCredentials,
   rebuildPluginNativeDeps,
@@ -56,6 +57,7 @@ import { applyClaudeSkipPermissions, applyCodexSkipPermissions } from '../lib/sk
 import { providerForCreate } from '../provider/registry.js';
 import { cloudAgentStartDetached } from './_cloud-attach.js';
 import { spawnQueuedOpenTerminal } from '../terminal/queue-open.js';
+import { resolvePortlessNonInteractive } from '../portless-prompt.js';
 import { buildSetupInitialPrompt } from '../wizard.js';
 
 /**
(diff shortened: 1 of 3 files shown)


madarco:
Integrated into `nightly` (fast-forwarded `109ca1238..80dee039f`), following the nightly integration flow. Closing this PR against `main` — the fix now rides on `nightly` alongside the other in-flight features.

PR #184, opened 2026-07-10
fix(git-shim): pass local/file:// clones through to real git

madarco:
## Problem

`agentbox create --provider <any cloud>` fails immediately when agentbox itself runs **inside an agentbox box** (the documented agentbox-in-agentbox workflow):

```
agentbox git shim: unsupported flag '--no-checkout' for 'git clone'. Allowed: --branch, --depth
```

The in-box `git` shim (`packages/sandbox-docker/scripts/git-shim`, installed at `/usr/local/bin/git` ahead of the real git, and shipped verbatim by **all five providers**) intercepts *every* `git clone` and routes it to `agentbox-ctl git clone` → host relay.

But agentbox's own cloud workspace seeding runs a purely local, credential-free clone — `runShallowClone` in `packages/sandbox-cloud/src/sync/workspace-seed.ts`:

```
git clone --no-checkout --quiet [--depth=N] file:///workspace <tmpdir>
```

It calls `execa('git', ...)`, which resolves through PATH and hits the shim.

## Why not just allow `--no-checkout`

Because relaying it would still be wrong. `agentbox-ctl git clone`'s own help says *"Clone a **github** repo into the box. Host runs `git clone` with its creds."* A `file://` clone touches no credentials and has no business going to the host.

So: identify the clone source **before** the flag gate, and `exec` the real git when it's a filesystem path or `file://` URL. Remote URLs and `owner/name` shorthand keep the existing gate, unchanged and fail-closed.

This is not a privilege boundary — `/usr/bin/git` is already on PATH behind the shim, so a local clone was never something the shim prevented. Only the source scan is new; `--branch`/`--depth` value-consumption is handled so a branch name can't be mistaken for the source.

## Verification

- `packages/ctl` suite: **326/326** pass with the fixed shim as PATH's git. 5 new tests cover file:// fall-through with `--no-checkout` intact, absolute-path fall-through, `--branch <name>` not desyncing the scan, remote `--no-checkout` still rejected, and `owner/name` still relayed.
- **Live**: with only the fixed shim on PATH (no bypass), `agentbox create --provider vercel` from inside a box now succeeds and seeds `/workspace` on the correct `agentbox/<name>` branch at the right HEAD. Box destroyed after.
- Side effect: `packages/ctl/test/bootstrap.test.ts` (which shells out to `git clone --bare`) currently **fails inside a box** on nightly because of the same shim. This fixes that too.

## Scope

One canonical file — every provider (docker / daytona / hetzner / vercel / e2b) installs this same script, so the fix lands everywhere at once. Takes effect for existing boxes only after the base images are re-baked (`agentbox prepare --provider <name> --force`).

## Not fixed here

The docker provider separately fails in-box: it bind-mounts the host repo's `.git` at its identical absolute path (`create.ts:724`), which inside a box is `/workspace/.git` and collides with the container's own `/workspace` worktree mount. `create` exits 0 with a silently broken git. Confirmed via control test (same provider, repo at a non-`/workspace` path → works). Left for a separate change.

https://claude.ai/code/session_01K3PimZHyJjxVhuCfbHoZqX

---

> [!NOTE]
> **Medium Risk**
> Changes routing for `git clone` in a credential-sensitive shim, but only widens passthrough for local paths; remote URLs remain gated and fail-closed on disallowed flags.
> 
> **Overview**
> Fixes **agentbox-in-agentbox** cloud creates that die on workspace seeding: the box `git` shim used to send every `git clone` to `agentbox-ctl` and reject flags like `--no-checkout`, but seeding runs a local `file:///workspace` clone that should never hit the host relay.
> 
> **`git-shim`** now finds the clone source before the flag whitelist. Sources that look like `file://`, absolute paths, or relative/`~` paths **`exec` real git** with the full argv (including `--no-checkout`). GitHub URLs, `https://…`, and `owner/name` still go through **`agentbox-ctl`** with the same strict flags; remote `--no-checkout` stays rejected. **`--branch` / `--depth`** values are skipped when picking the source so a branch name is not treated as the URL.
> 
> **`gh-and-shims.test.ts`** adds `makeRealGitStub`, optional `extraEnv` on `runShim`, and five tests for fall-through vs relay behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e23996bcd9fdfdb1b3d67305b8c0bf4a03b40e61. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ctl/test/gh-and-shims.test.ts
+++ b/packages/ctl/test/gh-and-shims.test.ts
@@ -627,6 +627,62 @@ describe('git-shim arg whitelist + passthrough', () => {
     }
   });
 
+  // Options we don't model also take a value. If the scan guessed flag arity,
+  // that value would read as the clone source and hand real git a REMOTE clone,
+  // slipping the relay and the whitelist. Classify every token instead.
+  it.each([
+    ['--reference', '/tmp/ref'],
+    ['--separate-git-dir', '/tmp/gitdir'],
+    ['--upload-pack', '/usr/bin/git-upload-pack'],
+    ['--reference-if-able', '/tmp/ref'],
+  ])('clone %s <path> before a remote url does not bypass the gate', (flag, value) => {
+    const env = makeStubShell();
+    try {
+      const realGit = makeRealGitStub(env);
+      const out = runShim(GIT_SHIM, ['clone', flag, value, 'https://github.com/x/y.git'], env, {
+        AGENTBOX_REAL_GIT_PATH: realGit,
+      });
+      expect(out.code).toBe(2);
+      expect(out.stderr).toMatch(/unsupported flag/);
+      expect(out.stdout).not.toContain('REAL_GIT:');
+      expect(out.stdout).not.toContain('STUB:');
+    } finally {
+      env.cleanup();
+    }
+  });
+
+  // scp-style and ssh:// remotes must not be mistaken for local paths either.
+  it.each([['git@github.com:owner/name.git'], ['ssh://git@host/owner/name.git']])(
+    'clone %s is treated as remote, not local',
+    (url) => {
+      const env = makeStubShell();
+      try {
+        const realGit = makeRealGitStub(env);
+        const out = runShim(GIT_SHIM, ['clone', url, '/tmp/dest'], env, {
+          AGENTBOX_REAL_GIT_PATH: realGit,
+        });
+        expect(out.stdout).not.toContain('REAL_GIT:');
+      } finally {
+        env.cleanup();
+      }
+    },
+  );
+
+  // Ambiguous/unrecognized source (no local token at all) ⇒ gate, not real git.
+  it('clone of a bare remote name falls to the gate, not real git', () => {
+    const env = makeStubShell();
+    try {
+      const realGit = makeRealGitStub(env);
+      const out = runShim(GIT_SHIM, ['clone', 'myremote'], env, {
+        AGENTBOX_REAL_GIT_PATH: realGit,
+      });
+      expect(out.stdout).not.toContain('REAL_GIT:');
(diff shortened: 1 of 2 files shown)


madarco:
**Bugbot is right — fixed in 5b7108f2.**

Confirmed the bypass on the previous commit:

```
git clone --reference /tmp/ref https://github.com/x/y.git
  nightly:  agentbox git shim: unsupported flag '--reference'   # loud
  e23996bc: Cloning into 'y'...                                 # silent bypass
```

`--reference` takes a value the scan did not model, so `/tmp/ref` read as the clone source, matched the local pattern, and exec'd real git with the full argv — remote URL included — skipping both the relay and the strict whitelist. Same for `--separate-git-dir`, `--upload-pack`, `--reference-if-able`.

Not a credential leak (the box holds no git creds, and `/usr/bin/git` is already on PATH behind the shim), but it converted a clear rejection into a confusing one. That is precisely the fail-open an argv gate must never do.

**Fix:** stop guessing flag arity. Classify *every* non-flag token — flag values included — and fall through only when something is unmistakably local **and** nothing looks remote (`scheme://`, scp-style `host:path` / `user@host:path`, or the bare `owner/name` shorthand `ctl` accepts). Unknown or unclassifiable sources take the gate. Fail-closed by default, so an unmodelled flag can no longer desync the parse.

**Regression tests added** for `--reference` / `--separate-git-dir` / `--upload-pack` / `--reference-if-able` before a remote, scp-style + `ssh://` remotes, and an unclassifiable bare source. Local seeding (`--no-checkout --quiet --depth=N [--branch <b>] file:///workspace`) and bare local clones still fall through. Suite: 333/333.

PR #185, opened 2026-07-10
fix(git-shim): allow harmless read-only flags; route pull --ff-only to the merge

madarco:
## Context

User bug report:

> The git shim rejects common read-only fetch flags. `git fetch --quiet` failed with `unsupported flag`, which silently broke the first rebake's checkout under `set -e`. → Allow `--quiet` and other harmless read-only fetch flags.

Reproduced verbatim. The rejection comes from the bash PATH shim installed at `/usr/local/bin/git` (`packages/sandbox-docker/scripts/git-shim`), not from `agentbox-ctl`. It intercepts the four credential-touching ops and enforced a per-op allowlist of **exactly one flag each**: `--force-with-lease`, `--ff-only`, `--prune`.

## Also fixes: `git pull --ff-only` was broken

Found while confirming the report. `--ff-only` is pull's *only* allowed flag, and it never worked. The shim forwards flags after a `--` separator, so commander treats them as passthrough positionals, leaves `opts.ffOnly` unset, and the flag gets appended to the host-built `git fetch <remote> <branch>` — which rejects it:

```
$ /usr/bin/git fetch --ff-only
error: unknown option `ff-only'
```

`partitionPullArgs` now routes `--ff-only` to the local merge, fetch flags to the fetch RPC, and `--quiet` to both.

## The change

| Op | Now allowed |
| --- | --- |
| `push` | `--force-with-lease`, `--tags`, `--dry-run` |
| `pull` | `--ff-only`, `--prune`/`-p`, `--tags`/`--no-tags` |
| `fetch` | `--prune`/`-p`, `--prune-tags`, `--tags`/`--no-tags`, `--force`/`-f`, `--dry-run` |

All three additionally take `--quiet`/`-q`, `--verbose`/`-v`, `--progress`/`--no-progress`. `clone` is unchanged.

## The gate stays fail-closed

Every push/pull/fetch flag must be **value-less**, and that is structural rather than a convention: the glued form (`--upload-pack=/bin/sh`) fails the exact-match regex, and the split form (`--depth 1`) has its value caught by `refuse_positionals`. That excludes command-executing flags (`--upload-pack=`, `--receive-pack=`, `--exec=`, `-o`) for free. Both halves are now pinned with tests.

Refused on purpose, with a comment saying why: `--delete`/`--mirror`, `push --prune` (destructive on the remote), `push --force` (use `--force-with-lease`), and `--all` (git rejects it alongside the relay's positional remote+branch).

Worth stating plainly: the shim was never the security boundary — `agentbox-ctl git` accepts unknown options, so an in-box process can bypass it. The host approval gate is what actually holds. The shim's job is preventing confusing failures.

## Verification

- 101 unit tests pass, including 7 new shim cases (`fetch --quiet`, `fetch --prune --quiet`, `pull --ff-only`, `push --quiet --dry-run`, and the value-consuming/destructive rejections) and 4 `partitionPullArgs` cases. There were previously **no** `fetch` or `pull` cases at all.
- Rebuilt the box image and confirmed the new shim is baked in.
- Drove the real path end-to-end (shim → ctl → stub relay → local merge) against a repo one commit behind its upstream: the relay received `git.fetch args: []` for `pull --ff-only` (previously `["--ff-only"]`), and the merge genuinely fast-forwarded — HEAD advanced and the new file appeared. `git fetch --quiet` now exits 0 silently.

One existing test, `packages/ctl/test/bootstrap.test.ts`, fails inside a box both before and after this change: it doesn't neutralize the global git config, so it inherits the host `~/.gitconfig`'s `commit.gpgsign` pointing at a nonexistent key path. Untouched here.

## Rollout note

The shim is **baked** into every image and snapshot, never pushed at create time. Docker rebuilds off the build-context fingerprint, but existing cloud boxes keep the old shim until the user re-runs `agentbox prepare --provider <name>`.

## Docs

`docs/host-relay.md` and `apps/web/content/docs/sync-and-git.mdx` now document the allowlist, the value-less invariant, and the deliberately-refused set. Neither enumerated it before.

https://claude.ai/code/session_0138cdMrFoxoYWSQCBAMsgUX

---

> [!NOTE]
> **Medium Risk**
> Changes the in-box git relay path agents rely on for fetch/pull/push; behavior is mostly additive but pull flag routing affects merge semantics, and updated shims only apply after image/snapshot rebuild.
> 
> **Overview**
> Fixes in-box `git` failing on common flags (e.g. **`git fetch --quiet`** under `set -e`) by **widening the shim allowlist** for `push`/`pull`/`fetch` with shared verbosity flags (`--quiet`, `--verbose`, `--progress`, etc.) and more read-only options (`--dry-run`, `--tags` on push, prune/tags on pull/fetch). Destructive or value-taking flags stay blocked via the existing **value-less** allowlist rules.
> 
> **`git pull --ff-only`** is fixed in **`agentbox-ctl`**: new **`partitionPullArgs`** sends `--ff-only` to the **local merge** instead of the relay **`git fetch`**, which used to reject it when the shim forwarded flags after `--`.
> 
> Docs in **`sync-and-git.mdx`** and **`host-relay.md`** now list the shim flags and rollout note (image/snapshot rebake). Tests cover shim allow/deny cases and pull arg splitting; **`bootstrap.test.ts`** avoids the box git shim and host GPG config when exercising real `git`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5bc47d85ad786941cae75b2a71054f3564fae3c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #186, opened 2026-07-10
fix(login): drive agent sign-in under a pty, prompt on the host

madarco:
## The bug

A user on **kitty** reported `agentbox claude login` hanging: once the browser opened, the "Paste code here" prompt stopped accepting input. `--headless` + `--code` worked, but it isn't the default.

The cause is structural, not terminal-specific. All three agent `login` subcommands did `spawnSync('docker', ['run','-it',…], { stdio: 'inherit' })`, handing the user's terminal straight to the agent's **own in-container TUI**. That TUI negotiates raw-mode and enhanced-keyboard sequences with whatever emulator it finds, and we control none of it. Kitty (CSI-u keyboard protocol) is just the emulator that exposed it.

The reporter suggested "default to headless when not in iTerm/tmux". A terminal allowlist is the wrong axis — it hardcodes today's known-good set and breaks on the next emulator.

## The fix

Never give the container's TUI the user's terminal. The login container now runs under **node-pty**, its raw stream goes to the command log, and we reproduce the interaction with **our own clack prompts on the host** — identical in every terminal. This is the new TTY default (`guided`) for claude, codex, and opencode.

`--interactive` opts back into the passthrough (also the automatic fallback when the optional node-pty prebuild is absent). `--headless` / no-TTY is unchanged.

The pty core already existed for claude's `--headless` worker; this generalizes it:

| file | role |
| --- | --- |
| `lib/agent-login-specs.ts` | pure per-agent prompt detectors, unit-tested against real captured transcripts |
| `lib/agent-login-run.ts` | the pty loop, agent-agnostic |
| `lib/agent-login-bindings.ts` | per-agent docker argv + `verify`/`finalize` |
| `lib/guided-login.ts` | the host-side clack prompts |
| `lib/claude-login-run.ts` | now a thin wrapper — the `--headless` worker and the hub create-job worker are untouched |

The first-run sign-in offers (plain `agentbox claude`, hit far more often than the subcommand) route through the same seam.

## Per agent, driven by an empirical capture phase

Captured by driving the real binaries under a pty before designing (transcripts in `docs/agent-login-guided-plan.md`). Two findings reshaped the plan:

- **codex** `--device-auth` prints a URL + one-time code, then polls — **no keystroke ever**. So guided codex works with no TTY at all (`codex login` no longer requires one). Its other methods (`--api-key`) never print a URL, so they skip guided rather than waiting out the 60s URL timeout.
- **opencode** is a per-provider prompt **tree**, not one prompt: `-p anthropic` asks for an API key, `-p github-copilot` opens a nested "Select GitHub deployment type". And `opencode models` only lists providers you already have creds for, so the picker can't be enumerated without duplicating opencode's models.dev registry. Guided opencode is therefore deliberately **bounded**: ask for the provider id on the host, pass `--provider` to skip the picker, drive the API-key / OAuth shapes, and fall back to the passthrough on anything else.

## Verification

`pnpm build`, `pnpm lint`, 743 tests pass (13 new extractor tests fed the real transcripts + a widened `selectLoginMode` matrix).

Beyond that, the real guided driver was driven through the PTY harness against the **real** `claude` / `codex` / `opencode` binaries, with only the docker layer replaced by an argv shim:

- **claude** — our "Paste the code from the browser" prompt renders on the host; a submitted code is written back into the pty; a **rejected** code surfaces `the code was not accepted — paste a fresh one` and re-prompts. No Claude TUI frames on screen; the raw stream lands in `~/.agentbox/logs/claude-login.log`.
- **codex** — device URL + one-time code print from the host, then a waiting spinner. No keystroke needed.
- **opencode** — the masked API-key prompt round-trips (opencode persisted the key to `auth.json`); `-p github-copilot` is detected as unsupported and falls back; `-p nope` reports `unknown provider "nope"`. The key appears **zero** times in the command log.

## Note for review

`guided-login.ts` imports `@clack/prompts` directly rather than the repo's `lib/prompt.js` wrapper. The wrapper `process.exit`s on Ctrl+C, which would strand the login container; we need the cancel symbol back to abort the pty first. Commented at the import.

Docs updated in the same change: `apps/web/content/docs/cli.mdx`, `run-an-agent.mdx`, plus `docs/agent-login-guided-plan.md`.

https://claude.ai/code/session_01AikKYvWcp1gV4vb2uqATTA

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth/credential capture and login UX across three agents; mistakes could strand containers or mishandle secrets, though guided mode avoids echoing API keys and preserves headless Claude behavior.
> 
> **Overview**
> Fixes broken agent login on terminals like **kitty** by making **guided sign-in** the default: login runs in a throwaway Docker container under **node-pty**, raw output goes to logs, and **host clack prompts** collect OAuth codes, device codes, or API keys instead of handing the terminal to each agent’s in-container TUI.
> 
> Adds shared **`agent-login-specs`**, **`agent-login-run`**, **`agent-login-bindings`**, and **`guided-login`**, and refactors **`claude-login-run`** to wrap the generic pty loop (headless worker / queue paths unchanged). **`selectLoginMode`** gains **`guided`**; **`--interactive`** restores the legacy `stdio: 'inherit'` passthrough (also used when node-pty is missing).
> 
> **Claude**, **Codex**, and **OpenCode** `login` and first-run sign-in flows route through **`signInTo*`** helpers. Codex guided mode targets **`--device-auth`** only and can run without a TTY; other Codex login args skip guided. OpenCode asks for **`--provider`** on the host when needed and falls back on unsupported prompt shapes. Docs and unit tests for prompt detectors are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f06a59ef62d71f4b0d889d2172fff24be9d879a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Bugbot's finding is valid — fixed in b6c12f40c.

`rejectionMessage(need)` now derives the retry hint from the pending `LoginNeed`, so a rejected opencode API key says `that API key was not accepted — try another` instead of the OAuth-code wording. Unit-tested both branches.

Also pushed c5eab1169, unrelated to this branch: `pnpm typecheck` has been red on nightly since 63c2795b5 (the `scopeFlag === 'none'` guard narrows the flag, so the inferred `scope` couldn't hold the `'none'` the picker returns). It was failing CI on every PR.

PR #201, opened 2026-07-12
fix(update): notice a tray-only release

madarco:
## Problem

The menu-bar app ships on its own cadence, so a **tray-only release bumps no CLI version** — and nothing in the system was watching the app's version. A tray-only release reached nobody automatically:

- **Tray side** — `UpdateChecker` fetched npm's `latest` **CLI** and compared it against `agentbox --version`. `CFBundleShortVersionString` was never compared to anything, so the app had no self-knowledge of whether it was stale. It reported "up to date" (true of the CLI) while running old code.
- **CLI side** — the daily probe already fetched and cached the published tray zip sha (`trayLatestSha`), and `maybeUpdateTray()` would have installed it, but the startup refresh prompt only fires on a **CLI version change**. So the cached sha was never acted on.

Net: v0.1.9 of the app (the DigitalOcean token fix) sat unnoticed on a test machine; only an explicit `agentbox install app` / `self-update` picked it up.

## Change (CLI half — the tray half is in the private tray repo)

- **Prompt on tray staleness too**, even when the CLI version is unchanged (`shouldPromptTrayUpdate`). Reads the daily cache — no new network on the command path.
- **Only the app is stale, so only the app is touched.** This deliberately does *not* route through `runPostUpdateRefresh()`, which also drops the box image and bounces the relay — disproportionate side effects for "a new menu-bar app exists". It calls `maybeUpdateTray()` alone.
- **Anti-nag**: declining stamps `trayDeclinedSha`, or the prompt re-fires on every command. A newer release (different sha) asks again.
- **Publish the app version as data** (`version.json` on the tray release) and cache it (`trayLatestVersion`) so the prompt can name it. Until now the version existed only as prose in the release title — which is precisely why the app itself had no way to tell whether it was stale. The `.sha256` sidecars stay: they still drive the install decision, and older CLIs depend on them.

## Verification

Driven end-to-end through the PTY harness against the live release (not just compiled):

- Stale tray + current CLI → prompt fires, naming the version.
- Decline → `trayDeclinedSha` stamped → next command **silent** (0 prompt occurrences). Clear the stamp → asks again.
- Accept → real download + install: `/Applications/AgentBox.app` went **0.1.7 → 0.1.9**, `traySha` restamped.
- Everything current → **no** false nudge (guards against an always-stale comparator).
- Full suite green (786 passed), lint + typecheck clean.

## Note

The tray-side check only takes effect from the app release that ships it (a stale tray can't run a checker it doesn't have). This CLI half works on every existing install immediately — it's the channel that actually closes the gap.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to optional interactive update prompts and background cache merging; install still keyed on sha sidecar with extensive unit test coverage.
> 
> **Overview**
> **Surfaces tray-only menu-bar app updates** when the CLI version is unchanged: interactive commands now prompt to install a stale tray build using cached daily probe data (`trayLatestSha`), calling `maybeUpdateTray()` only—**not** full `runPostUpdateRefresh()`—so the box image and relay are untouched.
> 
> Adds **`shouldPromptTrayUpdate`** (wraps existing `decideTrayUpdate` plus decline handling), **`trayDeclinedSha`** in `update-state.json` so declining one release does not nag on every command, and **`version.json`** fetch/cache (`trayLatestVersion`) so the prompt can show the published app version. **`mergeRemoteCheck`** keeps sha and version aligned when merging partial daily probes (drops stale version if sha moved without a new manifest).
> 
> Docs note separate tray vs CLI versioning and the new prompt behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36a6a26057bdf38009da3235c17b21d41beab385. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #203, opened 2026-07-12
fix(self-update): restart the hub, not just the relay

madarco:
## The report

Destroying a box failed with:

```
Cannot find module '/…/apps/hub/dist-standalone/apps/hub/dist-XULFTIMT.js'
imported from /…/apps/hub/dist-standalone/apps/hub/server.js
```

## Real bug, not a test artifact

The hub and the relay are **separate processes with separate pid files** (`hub.pid`, `relay.pid`), and `runPostUpdateRefresh` only ever restarted the **relay** — it never mentioned the hub.

An update replaces the installed package directory *underneath* whatever is still running. A hub left alive across `agentbox self-update` (which runs `npm install -g`) therefore keeps executing a bundle whose files no longer exist:

- **Lazily-imported chunks vanish.** The bundle's chunk names are content-hashed, so a rebuild writes `dist-<newhash>.js` and deletes the old ones. The running process only resolves them when a code path first needs one — which is why this surfaced on `destroy` rather than at startup.
- **Its cwd is deleted**, so every worker it spawns dies on `process.cwd()` (the `uv_cwd` crash fixed in #202 — same family, different process).

It stayed broken until someone restarted the hub by hand, which is not something a user would know to do.

## The fix

Restart the hub when it is the one running. The hub serves the relay on the same port, so the relay branch is left as-is for a bare-relay host — starting a second process there would just fight over 8787.

## Verification

Ran the real refresh against a live hub:

```
●  stopped hub (pid 32525)
●  hub back up on https://agentbox.localhost
```

pid changed 32525 → 41650, `/api/v1/health` reports the current version, and `/api/v1/boxes` (a route that had been failing on the missing chunk) serves again. Full suite green, lint + typecheck clean.

Companion to #202 — that one stopped a *poisoned hub* from killing the workers it spawned; this stops the hub from being poisoned in the first place.

---

> [!NOTE]
> **Medium Risk**
> Touches long-lived process lifecycle on self-update; logic is gated to avoid promoting relay-only hosts to a hub, but mis-detection could still cause a brief port conflict or wrong restart path.
> 
> **Overview**
> **Post-update refresh** now restarts the **hub** when that is what is actually running, instead of always cycling only the relay.
> 
> After `npm install -g`, the old hub process could keep running from a deleted install tree, which broke lazy-loaded bundle chunks and worker `cwd` until a manual restart. The refresh path uses `getHubStatus()` and branches on **`ui` or a live hub pid** (not bare `/healthz` `running`, which relay-only hosts also satisfy): **stopHub → ensureHub** for hub hosts, **stopRelay → ensureRelay** unchanged for relay-only hosts so nothing fights over the shared port.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 355ca08989d63b161f4265419780f10ccdd97c66. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/lib/post-update-refresh.ts
+++ b/apps/cli/src/lib/post-update-refresh.ts
@@ -14,8 +14,11 @@
 
 import {
   DEFAULT_BOX_IMAGE,
+  ensureHub,
   ensureRelay,
+  getHubStatus,
   removeImage,
+  stopHub,
   stopRelay,
 } from '@agentbox/sandbox-docker';
 import { installHostSkills } from '../commands/install.js';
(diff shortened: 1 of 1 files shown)


madarco:
bugbot run

PR #207, opened 2026-07-12
feat(provider): remote-docker — run boxes on a machine you own, over SSH

madarco:
Run a box as a container on a machine you already own — the workstation under the desk, a team server, a spare Mac — instead of the laptop or a cloud.

```bash
agentbox docker:buildbox claude
```

No account, no token, no bill: it connects **as you**, over your own `~/.ssh/config`. If `ssh <host> docker version` works, you can run boxes there.

## Why it's a CloudBackend and not a second docker provider

The docker provider's load-bearing trick is bind-mounting the host's `.git/` into the container, so an in-box commit lands in the host repo with zero sync. **A bind mount cannot cross a network** — and everything downstream of it is equally local-only (the ctl unix socket on a host path, `-p 127.0.0.1` on the *laptop's* loopback, `host.docker.internal` reaching the laptop's relay). Retargeting the docker provider with `DOCKER_HOST=ssh://` would retarget the `docker` calls and none of those assumptions; worse, the relay's git RPC would take the docker branch and try to push from a `.git` that isn't there.

The cloud scaffold already solved every one of them. So this implements the ~14 `CloudBackend` primitives over `docker`-on-SSH and inherits workspace seeding, the bridge relay, git-push-by-bundle, carry and credential sync unchanged. What stays docker-shaped: the image (a real Dockerfile build), checkpoints (`docker commit`), DinD, and `docker pause`.

## Design notes worth a look

- **Transport** — one OpenSSH ControlMaster per box; every docker call through `ssh <dest> bash -lc 'docker …'`. Deliberately *not* `DOCKER_HOST=ssh://`: docker's ssh transport needs `docker` on the **non-login** PATH, which is exactly where OrbStack and Colima break (a Mac remote is a supported case), and it reopens a connection per command.
- **Addressing** — the SSH destination is encoded into the sandbox id (`<dest>/<container>`). The relay resolves a backend from a bare handle with no access to config or box state, so the handle has to be self-describing; multi-host then falls out for free.
- **Image** — the ref *is* the build-context fingerprint (`agentbox/box:<sha>`), so "is this host prepared?" is answered by the engine rather than a local file that could disagree with it — and answered per host. Pull the multi-arch GHCR image, else stream the build context to `docker build -`.
- **`Provider.sshTarget`** (new seam) — `open`/`code` previously *reverse-engineered* the box's SSH target by parsing the `buildAttach` argv, which only works when the box IS the machine. A provider that knows its own target now says so. remote-docker's box sshd is reached by `ProxyJump`ing through the engine — plain OpenSSH, so VS Code Remote-SSH and sshfs work with no tunnel to keep alive.
- **SSH transport hoisted** into `@agentbox/sandbox-core` — hetzner and digitalocean carried near-verbatim copies that had already drifted; this would have been a third.

## Verified end-to-end

Against a real SSH-reached engine: `prepare` (GHCR miss → streamed build context → image on the remote), `create` (workspace seeded as a git clone, host's uncommitted **and** untracked changes carried over byte-identical), bootstrap, preview URLs, interactive attach (`ssh → docker exec → tmux`, driven in a PTY), `ssh <box>` via ProxyJump (incl. the ephemeral port being re-resolved across a stop/start), checkpoint, pause/resume, and destroy leaving no container or volume behind.

15 test suites + lint green; 38 new unit tests.

## Not done / known

- **Live-verify on a non-nested remote.** All of the above ran against an engine that is itself an AgentBox box (loopback ssh). A plain Linux server and a macOS/OrbStack remote each deserve a real run — `scripts/linux-dev-vm.sh up` exists for this. Tracked in `docs/remote-docker-backlog.md`.
- Ports are fixed at `docker run` (same constraint as Vercel); checkpoints are engine-local.

## Adjacent bug found (pre-existing, not fixed here)

`detectGitRepos` only accepts `.git` as a **directory**. When the source workspace is a git *worktree* — which `/workspace` is inside *any* AgentBox box — every cloud provider silently falls back to tarring the directory and the new box gets a dangling `.git` pointer. Hits daytona/hetzner/vercel/e2b/digitalocean equally. Worth its own fix.

---

> [!NOTE]
> **Medium Risk**
> Touches shared SSH/tunnel code used by hetzner and digitalocean, and adds a large new cloud lifecycle path; e2e was verified on a nested engine only, with plain Linux/macOS remotes still called out as backlog.
> 
> **Overview**
> Adds a **remote-docker** backend so boxes run as containers on a user-owned machine over SSH, using the existing cloud scaffold (workspace sync, bridge relay, git push via bundle) instead of local bind-mounted `.git`.
> 
> **CLI & addressing:** New `docker:<host>` / `remote-docker:<host>` provider specs via `parseProviderSpec`, sugar like `agentbox docker:buildbox claude`, `--remote-host`, and config `box.remoteDockerHost`. `cloudSizingProviderOptions` requires a remote host for this provider.
> 
> **New package `@agentbox/sandbox-remote-docker`:** Implements `CloudBackend` with one ControlMaster per box, remote `docker` via `ssh … bash -lc`, fingerprint-tagged images (GHCR pull or streamed build), `docker commit` checkpoints keyed as `<host>#<image>`, and sandbox ids that embed the SSH destination.
> 
> **SSH plumbing:** Shared `ssh-exec`, `SshTunnelManager`, and `ProxyJump` support in `@agentbox/sandbox-core`; hetzner/digitalocean re-export instead of duplicating. Optional **`Provider.sshTarget`** and `box.ssh.proxyJump` for `open` / `code` / VS Code Remote-SSH. remote-docker is included in persistent SSH, IDE, and sshfs provider lists.
> 
> **Hub, config, docs:** Provider registered in CLI, hub OpenAPI/validation, config schema (`remoteDockerHost`, size/image/checkpoint overrides), relay backend loader, prune list, and new user-facing docs plus backlog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 562ca8f3fbf04d1f23369743ae696d69ecc51655. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Integrated into nightly via merge commit ab53bd601 (nightly fast-forwarded to the feat/remote-docker tip, per the usual nightly integration flow).

PR #215, opened 2026-07-14
fix(config): unknown keys warn instead of failing the command

madarco:
## The bug

Both config parsers ship **inside artifacts that outlive the CLI that wrote the config**, and both were fail-closed on unknown keys:

- **`@agentbox/config` is inlined into every provider plugin** (`provider-sdk` sets `noExternal: [/^@agentbox\//]`, and `createCloudProvider` calls `loadEffectiveConfig` internally). A plugin pinned to an older SDK therefore parses the user's config with a **stale key registry**.
- **`@agentbox/ctl` is baked into the box image**, so a box created months ago parses an `agentbox.yaml` written against today's CLI.

So every config key we added was a **breaking change for already-published plugins and already-baked images**. This is not hypothetical — adding `box.daytonaVmBaseImage` (2026-07-12, 074ea83e0) killed an islo `create` on a plugin published two days earlier, *mid-flight, after the sandbox had already been provisioned*:

```
UserConfigError: ~/.agentbox/config.yaml.box: unknown key "daytonaVmBaseImage" (known: provider, hostSnapshot, …)
    at parseBranchObject (…/@madarco/agentbox-provider-sdk/dist/chunk-2BCOLDML.js:761)
    at loadEffectiveConfig (…) → at Object.create (…)
```

The user's CLI was fine. Only the plugin's bundled parser was stale.

## The fix

An unknown key is **skipped with a warning** — the rest of the config still applies and the command proceeds. Type errors, renamed keys, and `agentbox config set <bad-key>` all still fail loud: the registry stays authoritative for the keys it *does* know.

The host CLI surfaces the warnings (a process-wide sink installed once at startup, plus a `config` row in `agentbox doctor`); a plugin's bundled copy registers no sink and stays quiet — same code, right behavior on both sides.

## Two traps this had to avoid

- **`config set` would have silently deleted unknown keys.** It rewrites the file from the doc it read back, so once the parser started *dropping* unknown keys, the next `set` would erase exactly the forward-compat keys we just promised to preserve. `readExistingDoc` now round-trips raw YAML. Covered by a new test.
- **The JSON schemas stay `additionalProperties: false`.** They drive editor autocomplete, where flagging a typo as you type is the point. Parser and schema now deliberately disagree on unknown keys *only* — both schema-drift suites document the split rather than just loosening (`schemaOnly` fixtures on the ctl side, a new group on the config side).

## Verification

Unit: full monorepo green (`config` 139, `ctl` 353, `sandbox-cloud` 123, CLI 829, relay 287), lint clean.

Live — the real check:
1. **Reproduced** the failure: added a key no published SDK knows to `~/.agentbox/config.yaml` → `agentbox create --provider islo` (community plugin, real SDK bundle) died with `UserConfigError: unknown key`.
2. **Fixed**: dropped the rebuilt SDK `dist/` into that same plugin → the identical create succeeded, printed one warning, and the box worked end-to-end (git branch seeded, `agentbox.yaml` web service serving).
3. **In-box**: a docker box whose `agentbox.yaml` carries unknown top-level and task keys now boots, runs its tasks (`install` exit 0), and `agentbox-ctl validate` reports both keys and exits 0. On `nightly` today, that box refuses to start.
4. **Regression** on the common path: `config set` preserves an unknown key in the file; `config set box.typoKey` still errors; `doctor` lists the key as a warn row.

## Follow-up (not in this PR)

`@madarco/agentbox-provider-sdk` needs a **minor bump + republish** for this to reach existing plugins — their bundled copy is the broken thing. Additive only (`LoadedConfig.warnings`, `setConfigWarningSink`), so `SDK_API_VERSION` stays `2`.

https://claude.ai/code/session_014Lp4fP5ZaqLzrd3LXksmkU

---

> [!NOTE]
> **Medium Risk**
> Touches config parsing on every command and box boot, but behavior change is narrowly scoped to unknown keys; type errors and explicit `config set` on bad keys still fail loud.
> 
> **Overview**
> Fixes forward-compat breaks when **older parsers** (provider SDK bundles, baked `agentbox-ctl`) hit config keys from a **newer CLI** — e.g. `create` dying on an unknown `box.*` key after the sandbox was already provisioned.
> 
> **`@agentbox/config`:** Unknown sections/leaves call `ParseOptions.onWarning` and are **ignored**; wrong types and renamed keys still throw. `loadEffectiveConfig` returns **`warnings`** and optional **`setConfigWarningSink`** (CLI registers once at startup with de-duplication; plugins stay silent). **`config set`/`unset`** read **raw YAML** so unrecognized keys are not dropped on rewrite.
> 
> **`@agentbox/ctl`:** `rejectUnknownKeys` → **`warnUnknownKeys`**; **`CtlConfig.warnings`** logged at daemon start and printed by **`validate`** (exit 0).
> 
> **CLI:** `agentbox doctor` adds **config** warn rows for unknown keys. Docs/schema-drift tests document the deliberate split: **JSON Schema stays strict** for editors; runtime parsers warn only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c23623b79964c8ae12614819c71629ccf09ac4cb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Added the SDK bump: **`@madarco/agentbox-provider-sdk` 2.3.0 → 2.4.0** (519e85ff9).

This is the part that actually reaches existing plugins — they bundle their own copy of `@agentbox/config`, so *their* copy is the one that fails closed on an unknown key. A user upgrading only the CLI stays broken until the plugin picks up a new SDK.

Additive surface (`LoadedConfig.warnings`, `setConfigWarningSink`), so it's a minor bump and `SDK_API_VERSION` stays `2` — existing plugins remain loadable. Publish gate is green: `pnpm --filter @madarco/agentbox-provider-sdk pack:test` → 25 exports present, SDK_API_VERSION=2, and the tolerant parser is confirmed present in the packed bundle.

`npm publish` is still a manual step for you (per the repo's release flow).

PR #250, opened 2026-07-24
refactor(cli): one `hub` command group, then close the remaining remote-hub gaps

madarco:
## Phase 5 — rename `control-plane *` → `hub *`

Completes the final phase of `docs/hub-api-consolidation-plan.md`. AgentBox is unreleased, so this is a **clean rename** (no aliases): the hidden top-level `agentbox control-plane` group is gone, and its subcommands are folded into the single **`agentbox hub`** group.

### What changed
- **One `hub` group.** `hub setup`/`deploy`/`set-url`/`unset-url`/`add`/`worker`/`credentials`/`secrets`/`project`/`custody`/`boxes`/`prompts` are now `agentbox hub` subcommands, all surfaced in `agentbox hub --help`. `start` stays the default (bare `agentbox hub`).
- **`hub status` unified by target.** The one hard collision: `agentbox hub status` now reports the configured **remote control box** (reachability + box/event counts) when one is set (`--url` forces remote), else the **local hub process** (pid/port). Extracted `probeControlPlaneStatus` for the remote path.
- **All user-facing command strings** (help/errors/arg descriptions) across `apps/cli` + the provider packages updated to `agentbox hub …`.
- **Docs:** `deployed-hub`/`cli`/`configuration`/`api` mdx + internal plans; the consolidation plan's Phase 5 marked done with the remaining items documented as next.

### Deliberately left (not the CLI verb)
The on-disk `~/.agentbox/control-plane/` dir + `control-plane.env`, the `relay.controlPlaneUrl` config key, the `'control-plane'` `SyncTopology` enum, and the `apps/cli/src/control-plane/` source folder.

### Documented as next (not in this PR)
Web-UI create seed overlay, and `ls -g` / `create --via-hub` onto `/api/v1`. (Destroy-reap was listed here too — it is now done below.)

---

## Then: closing the gaps that audit found

The rename added an appendix to `docs/control-box-plan.md` auditing, per command, whether it talks to the control box and where it *should* run. Every row was re-verified against the code, which turned up three things:

**Two listed gaps were already closed.** `--via-hub` for `claude`/`codex`/`opencode` (all route through `resolveCreateRouting`; `fork` stays local **by design** — it always forwards `--resume`/`--continue`, and session teleport copies host state at create time, which the VPS worker can't do), and the web UI driving PC-registered boxes (reverse adoption).

**One real defect wasn't listed at all** — the attach footer, filed under a plain "Local" row for `attach`/`shell`.

The common shape of what remained: the PC kept talking to *itself* about a box that lives on the control box. The fixes are client-side; the plane's routes already accepted a non-loopback caller bearing the admin token (`adminGateAllows`).

- **`06fcc8a4` reap on destroy/prune.** `reapBox()` and its `DELETE /remote/boxes/:id` route both existed with **zero call sites** outside a test, so PC-destroyed boxes lingered as ghosts in `ls`, the web UI and the tray. Best-effort + bounded: an unreachable control box warns and names `hub boxes rm`, never fails the local destroy. Also fixes a pre-existing bug this uncovered — `pruneBoxes` judged every record by `docker inspect`, and a cloud box's container is the synthetic `cloud:<sandboxId>`, so a plain `agentbox prune` **deleted the state record of every live cloud box** (and with `--all`, its per-box dir holding the private SSH key). Guarded on provider, as `listBoxes` already does.
- **`571a5da7` approvals follow the box's own relay.** The attach footer, `agent approvals`/`approve` and the dashboard marker were all pinned to `127.0.0.1:8787`, so a hub box's approvals were invisible and unanswerable from the CLI — the tray or web UI was the only way to unblock one. `resolveBoxPromptSource` resolves each box's plane from its own record (`cloud.controlPlaneUrl`, which outlives a config change). `approve` takes an id with no box, so it tries local then the control box (ids are UUIDs, so unambiguous).
- **`08454b76` dashboard lists hub boxes.** `mergeHubBoxes` had exactly one consumer (`list.ts`); the shared `listBoxesMerged()` now backs both, and selecting an un-adopted row adopts it first so the panels/actions have a local record.
- **`d32f0d8a` hub create-queue visible from the PC.** Background `-i` cloud runs are enqueued on the plane, but `queue list` read only `~/.agentbox/queue/`. Adds `Store.listCreateJobs` (sqlite/postgres/memory + write-through, via the conformance suite so no backend drifts), `GET /remote/boxes`, `agentbox hub jobs list|show`, and a control-box block in `queue list`.
- **`3ff10c67` docs.** Corrects the audit appendix (closed / previously-mis-recorded / still-open) and documents the new surfaces, including that approval is not execution — an approved `git push` runs control-box-side, but a host-filesystem action still needs the laptop up.

### Verification (gap-closure commits)
`pnpm typecheck` + `pnpm lint` clean; 919 CLI and 380 relay tests pass, with new coverage for reap outcomes, the prompt-client auth header, the plane resolver, the merged listing, `listCreateJobs`, and the new route. Live-checked against stub control boxes: `hub jobs list/show`, `queue list` ±`--all`, the dashboard rendering a registration-only box in the sidebar (vs `[]` from the old source), and all three degradation paths — hub down (0.33s, exit 0), a pre-upgrade hub answering 405, and no admin token.

**Not yet done:** end-to-end against the live control box — a real create → destroy → confirm the registration is gone, and a real in-box `git push` approval answered from the attach footer.

### Verification
- `pnpm typecheck` (31/31) + `pnpm lint` (17/17) + build green; new `hub` command-tree test in `commands.test.ts` (fold has no drops/dupes; `status` has `--url`+`--json`; `boxes` lifecycle subs present); `help.test.ts` still green.
- **Live:** `agentbox hub --help` lists all 19 subs, `control-plane` gone from top level; `agentbox hub status` against the real control box reports `remote (reachable) — 3 boxes, 8 events` (+ clean `--json`); admin-token remote path reaches `/admin/registry` (200).

The rename touches no create/routing/worker logic — those paths are unchanged from the already-validated `-i`-on-hub (#248) and LFS (#249) PRs.

---

> [!NOTE]
> **Medium Risk**
> Touches destroy/prune lifecycle, approval routing, and dashboard SSE across local vs remote hubs; changes are mostly client-side with bounded fallbacks, but mistakes could leave stale registry rows or miss approvals until redeploy/token setup.
> 
> **Overview**
> **Consolidates hosted-hub CLI under `agentbox hub`** — the hidden `agentbox control-plane` top-level group is removed; `setup`, `deploy`, `boxes`, `prompts`, `credentials`, `custody`, `jobs`, etc. live on `hub`. **`hub status`** reports the configured remote control box (reachability + counts) when set, otherwise the local hub process.
> 
> **Fixes the PC talking to the wrong relay for hub-created boxes.** New `resolveBoxPromptSource` / `attachRelayOptions` route attach footers, `agent approvals`/`approve`, and dashboard SSE to each box’s control box (with admin bearer), not always `127.0.0.1:8787`. **`listBoxesMerged`** shares the hub registry merge between `ls` and **`dashboard`** (select → auto-adopt). **`destroy` / `prune`** call best-effort **`reapOnControlBox`** so registrations don’t ghost in `ls`/UI; **`prune`** also stops treating live cloud boxes as docker orphans. **`queue list`** and **`hub jobs list|show`** surface the control box’s create queue (where background `-i` cloud runs go).
> 
> Hub image/docs tweaks: **git-lfs** in the Dockerfile, **E2B attach-helper** in standalone build, **Daytona** deploy key fix (`DAYTONA_ORGANIZATION_ID` + endpoint overrides).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c07bdbf67108bf36c47c95e771726bfef5b602c6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Two follow-up commits added to this branch (deployed-hub fixes surfaced while live-validating the rename):

- **`2e56a78b` fix(hetzner): migrate Daytona JWT org id + endpoint overrides** — the deploy-time provider-secret allowlist carried `DAYTONA_ORG_ID`, but the real key is `DAYTONA_ORGANIZATION_ID`, so a JWT-mode Daytona setup landed on the control box with the token but no org id. Fixed + added endpoint/region keys + a drift sync test (`packages/sandbox-hetzner/test/provider-secret-keys.test.ts`).
- **`a5c98fd6` fix(hub): install git-lfs + stage the e2b attach-helper** — two packaging gaps that blocked a hub-routed cloud `-i` create on a fresh control box: the hub image lacked `git-lfs` (so #249's forced-HTTPS `lfs pull` failed → LFS repos broke the checkout), and the standalone build never staged `sandbox-e2b/dist/attach-helper.cjs` (so the box created but the `-i` agent couldn't start).

**Live-verified end to end:** a freshly deployed control box created an e2b box from an LFS repo, materialized the 512 KB LFS object over HTTPS, started claude detached, and the agent wrote the expected file (`hello.txt` = `hello`, ground-truth). Torn down + restored after.

Also filed (not fixed here — needs its own change): the PC's `box.claudeInstall` isn't migrated to the control box, so a PC that overrides it to `npm` bakes fingerprints the control box (default `native`) rejects — see the local-adoption backlog note.

madarco:
Five more commits: closing the remaining gaps from this doc's own **CLI remote-hub support audit** (the appendix added in this branch).

Re-verified every audit row against the code first. **Two listed gaps were already closed** — `--via-hub` for `claude`/`codex`/`opencode` (all route through `resolveCreateRouting`; `fork` stays local *by design*, since it always forwards `--resume`/`--continue` and session teleport copies host state at create time), and the web UI driving PC-registered boxes (reverse adoption). **One real defect wasn't listed at all**: the attach footer, which the audit filed under a plain "Local" row for `attach`/`shell`.

- **`06fcc8a4` reap on destroy/prune** — `reapBox()` and its `DELETE /remote/boxes/:id` route both existed with **zero call sites** outside a test, so PC-destroyed boxes lingered as ghosts in `ls`, the web UI and the tray. Also fixes a pre-existing bug this uncovered: `pruneBoxes` judged every record by `docker inspect`, and a cloud box's container is the synthetic `cloud:<sandboxId>` — so a plain `agentbox prune` **deleted the state record of every live cloud box**, and with `--all` its per-box dir holding the private SSH key. Guarded on provider, as `listBoxes` already does.
- **`571a5da7` approvals follow the box's own relay** — the attach footer, `agent approvals`/`approve` and the dashboard marker were all pinned to `127.0.0.1:8787`, so a hub box's approvals were invisible and unanswerable from the CLI; the tray/web UI was the only way to unblock one. Server side already worked (`adminGateAllows` admits a non-loopback caller with the admin bearer), so this is client-side only.
- **`08454b76` dashboard lists hub boxes** — `mergeHubBoxes` had one consumer (`list.ts`); shared `listBoxesMerged()` now backs both, and selecting an un-adopted row adopts it.
- **`d32f0d8a` hub create-queue visible from the PC** — background `-i` cloud runs are enqueued on the plane, but `queue list` read only `~/.agentbox/queue/`. Adds `Store.listCreateJobs` (all four backends, via the conformance suite), `GET /remote/boxes`, `agentbox hub jobs list|show`, and a control-box block in `queue list`.
- **`3ff10c67` docs** — corrects the audit appendix (closed / previously-mis-recorded / still-open) and documents the new surfaces.

**Verification:** typecheck + lint clean; 919 CLI and 380 relay tests pass, including new coverage for reap outcomes, the prompt-client auth header, the plane resolver, the merged listing, `listCreateJobs`, and the new route. Live-checked against stub control boxes: `hub jobs list/show`, `queue list` ±`--all`, the dashboard rendering a registration-only box in the sidebar (vs `[]` from the old source), and all three degradation paths — hub down (0.33s, exit 0), a pre-upgrade hub answering 405, and no admin token.

**Not yet done:** end-to-end against the live control box — a real create → destroy → confirm the registration is gone, and a real in-box `git push` approval answered from the attach footer.

madarco:
bugbot run

madarco:
All three bugbot findings fixed in `8587eefd` — all were real.

1. **Dashboard misses hub prompt relay** *(Medium)* — correct, and it defeated the point of pairing the dashboard merge with the approval routing. `relaySourceFor` looked the box up in local `listBoxes()`, so a row that exists **only** on the control box — exactly the one whose approvals are remote — resolved to nothing and fell back to loopback. Its marker would only light after adoption. Now resolved from the merged listing; `resolveBoxPlane` takes a narrower `PlaneAddressable` (`provider` + `cloud`) so a registration-only row resolves.
2. **Hub approve misreports missing token** *(Medium)* — correct. This is the same failure mode I explicitly handled in `approvals` and missed in `approve`: a control box we *can't ask* is not evidence the prompt is gone. It now distinguishes no-token from genuinely-resolved and exits non-zero.
3. **`listHubJobs` omits guard** *(Low)* — correct; `Array.isArray` guard, empty queue instead of iterating `undefined`.

Also fixed a **stream leak I introduced**, found self-reviewing the same code (not flagged by bugbot): the pending-slot sentinel in `syncPromptSubscriptions` was a shared singleton. If a box leaves and re-enters the list mid-resolve, two resolvers race; both match the shared sentinel, so the loser deletes the **winner's live stream** from the map — an SSE connection neither dispose nor teardown can close, re-subscribing on every poll after. Token is now per-attempt, and a bailing attempt only clears a slot still its own. The regression test fails against the old sentinel and passes with the fix (verified by reverting).

typecheck + lint clean; **922** CLI (+3) and 380 relay tests pass.

madarco:
bugbot run

madarco:
Bugbot round 2 — 3 more findings, all real, all fixed in `1403a49a`. (The three round-1 threads above are resolved by `eec77aab`; they re-anchored to the new commit but weren't re-reported — round 2 is review `4778779169`.)

1. **Dashboard destroy skips hub reap** *(Medium)* — correct, and an inconsistency I introduced: I wired the reap into `agentbox destroy` but not the TUI's `destroyBoxAction`, so destroying from the dashboard left exactly the ghost registration the CLI fix removes. Same reap now runs there.
2. **Hub job show needs full UUID** *(Medium)* — correct, and a genuine papercut: `jobs list` and the `queue list` block print an 8-char prefix, so the only id you can copy is one `show` would 404 on. `show` now tries the id verbatim, then resolves it as a prefix, reporting ambiguity rather than guessing.
3. **Adopt failure says unreachable** *(Low)* — correct; the placeholder asserted a cause it hadn't established (`tryAutoAdopt` also returns null for unknown-box / no-token). Reworded to point at `hub adopt`, which reports the real reason.

Verified against a stub control box: copied 8-char prefix resolves, ambiguous prefix lists candidates, full UUID unchanged, bogus id still not-found. lint clean, 922 CLI + 380 relay tests pass.

madarco:
bugbot run

madarco:
Bugbot round 3 — 2 findings, both real, fixed in `df2bd67d`.

1. **Remote relay error blocks TUI approvals** *(Medium)* — the important one, and a regression I introduced. `gatherApprovals` awaited the mailbox fetch before merging the in-TUI rows, and `fetchRelayApprovals` throws on any non-OK response. That was fine when it was a loopback call to a local daemon; making it a WAN request to the control box turned any blip there into a **command failure** that also hid the plan/question/permission rows — which come from `readBoxStatus` and are entirely independent of the relay. The relay half now degrades: rows reported as *missing*, never as absent; in-TUI rows still render; both text and `--json` exit 1 so a caller can distinguish "couldn't read the mailbox" from "nothing pending". `--json` keeps its array shape on stdout (orchestration reads it) with the warning on stderr.
2. **Dashboard omits reap failure notice** *(Medium)* — correct, and the mirror of the round-2 finding: I added the reap but discarded its result, so a box destroyed from the TUI could vanish locally while its hub registration survived, silently. `destroyBox` may now return a short note the confirmation flash appends — the TUI owns the screen, so an action's own `log.warn` would be invisible.

Verified against a control box that is down: `approvals` warns and still returns, exit 1 in both modes, `--json` stdout stays a parseable array. lint clean, 922 CLI + 380 relay tests pass.

Running total: **8 bugbot findings across 3 rounds, all real, all fixed** — plus one leak I found myself. Five of the eight are the same class: a fix applied to one path and not its sibling (`destroy` vs the TUI, `approvals` vs `approve`, list-prints-a-prefix vs show-wants-a-UUID). I've since audited both classes exhaustively — every cloud-destroy site and every `/admin/prompts` consumer — and they're consistent now.

madarco:
One more of the round-3 class, found by auditing for it rather than by review — `f0c61c06`.

**A dead prompt stream read as "no approvals".** `subscribePrompts` retries transient drops but bails **permanently** on a non-200, and `run.ts` never wired `onError`. Harmless while the URL was always loopback (a non-200 there is near-impossible). But a hub box now streams from the control box with an admin bearer, so a stale token answers **401 → the client closes for good → the attach footer sits silent for the entire session** while the box is blocked waiting on an approval. That is precisely the failure mode this PR exists to remove, reintroduced through a different door.

Now surfaced in the alert band (pointing at `hub prompts`) and passed to the caller's error sink so it's diagnosable after the fact.

The generalisable lesson across all of these: **this PR moved a set of calls from loopback to WAN, and every one of them had been written assuming it couldn't fail.** `fetchRelayApprovals` threw and took the command down; `subscribePrompts` died silently; `postAnswer` had a 3s timeout. Each was correct before and wrong after. I've now swept every call I relocated — that's the complete set.

madarco:
bugbot run

PR #251, opened 2026-07-25
fix(hub): a control box can use a bake made in either claudeInstall mode

madarco:
Fixes the two issues filed in `docs/local-adoption-plan.md` → Backlog. I verified both against the code first; one was valid as filed, the other was **narrower than filed**.

## Claim 1 — `box.claudeInstall` isn't migrated. **Valid.**

`prepare` folds the install mode into the stored fingerprint, but that key lives in the PC's `config.yaml` — a different file from `secrets.env`, so the deploy's provider-secret allowlist never carried it and a control box fell back to the built-in `native`. A PC set to `npm` pushed an `npm`-folded record its `native`-folded fingerprint never matched, so `hydratePreparedFromCustody` rejected it and **every cloud create failed with "run `agentbox prepare` first"** — on an identical build context.

**Root cause worth naming:** the record stores only `contextSha256`, **never the mode**, and the fold is one-way. So the backlog's suggested "have the worker honor the shared record's baked mode" isn't directly possible — the mode can only be *inferred*. There are exactly two modes and `native` is the identity fold, so one context hash yields both candidates: `matchClaudeInstallFingerprint` returns which mode a stored fingerprint came from. Adopting either is correct — the two bases differ only in how Claude Code was installed and are functionally equivalent, whereas rejecting means failing every create.

Applied to **both** consumers deliberately: hydrate alone fixes creates while `/settings` keeps reporting the just-adopted base as stale and demanding a re-bake.

Also migrates the key at deploy, so a control box that bakes its **own** base uses the operator's mode — which matters precisely here: `npm` exists because the native installer intermittently Cloudflare-403s datacenter egress IPs, and a control box is one. The migration **merges** into the VPS's existing config rather than regenerating it: the hub writes that same file itself (`box.remoteDockerHost` from Settings), so a fresh file would drop its keys on every redeploy.

## Claim 2 — control box can't self-bake. **Partially valid; the filed difficulty was stale.**

The backlog said this needs "the provider build machinery + queue integration on the control box". That machinery already ships: `prepareProvider` enqueues a `kind: 'prepare'` job, `POST /api/v1/providers/:id/prepare` exposes it, Settings has a bake button, and the **web-UI create is already two-phase**. Only the *worker* path is missing it. Corrected the doc rather than leave it to be re-scoped wrongly, and replaced "until the control box can self-bake" with a workaround that works today. No code change — out of scope by agreement.

## Live end-to-end (real control box, not stubs)

The control box already had the workaround applied by hand, so I **recreated the broken state** to prove anything:

| | |
|---|---|
| PC bake (`claudeInstall=npm`, this branch) | stored `1cec121c…` |
| Same context, native fold | `e75940ca…` |
| Control box config | `claudeInstall: native` ← un-migrated |
| Control box local record | removed ← fresh-control-box state |

Hub log:
```
prepared: adopted the shared e2b base from custody (no bake needed) (baked with claudeInstall=npm)
agentbox-hub: registered cloud box b05d1f437 (bakefix-e2e) with 1 worktree(s)
```

A hub configured for `native` adopted an **npm**-baked record and provisioned the box — the create that previously failed. `--via-hub` create returned `job done`. The merge was also run against the control box's **real** `config.yaml`: hub-written key preserved, migrated key applied, no duplication.

That surfaced one real property now pinned by a test: `sshExec` runs through execa, which strips the trailing newline, so the merge must accept a body without one.

## Notes

- 13 new unit tests; typecheck + lint clean; 925 CLI / 263 sandbox-core / 150 config / 85 hetzner pass.
- Stacked on #250 originally; cherry-picked onto `feat/local-adoption` after that merged, so this is only its own 3 commits.
- Found while setting up, not filed anywhere: `hydratePreparedFromCustody` early-returns when local prepared-state exists, so a control box holding a **stale** record never re-adopts from custody. Adjacent to this bug, not fixed here.

---

> [!NOTE]
> **Medium Risk**
> Touches control-plane deploy config migration and prepared-state adoption gates used on every hub cloud create; changes are narrow, well-tested, and best-effort on mismatch.
> 
> **Overview**
> Fixes control-box cloud creates failing with "run `agentbox prepare` first" when the PC baked with **`box.claudeInstall: npm`** but the VPS defaulted to **`native`** — the bake fingerprint folds install mode, and prepared records don't store which mode was used.
> 
> Adds **`matchClaudeInstallFingerprint`** to infer **native** vs **npm** from the stored hash and the current context's native fingerprint. **Custody hydration** and **hub provider base freshness** use it so shared bakes adopt correctly and **Settings** stops marking an adopted base stale.
> 
> **`hub deploy hetzner`** now **merges** `box.claudeInstall` from the PC's global config into the control box's `config.yaml` (via new **`mergeConfigYaml`**, preserving hub-written keys like `box.remoteDockerHost`). Docs and backlog notes are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84f374bd81d9065faeafe7437938b352742f5e0a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #258, opened 2026-07-26
feat(hub): install the control box from the npm package, keep --ref for source

madarco:
## Why

`agentbox hub deploy hetzner` cloned the monorepo onto the VPS and built 14 workspace packages plus the CLI there — **~2m40s of a ~4m45s deploy** — to produce an artifact the published package **already carries**: `runtime/hub/apps/hub/server.js`, the same standalone bundle `agentbox hub` spawns locally (`resolveHubServer`, staged by `stage-runtime.mjs`).

Building from a *ref* also couples the VPS artifact to a branch name rather than a version. That coupling is what produced the 502 fixed last week: `DEFAULT_DEPLOY_REF` was `main` while the CLI was a nightly, so `main`'s hub listened on `:3000` behind Postgres while the CLI wrote `reverse_proxy app:8787`.

## What changed

The VPS now runs `npm i -g @madarco/agentbox@<the deploying CLI's exact version>` in a small `node:22-slim` image, and **the host ships the compose stack**: `apps/hub/docker-compose.yml` verbatim plus `docker-compose.package.yml`, an override that replaces only the `build:` block. There is no repo on the VPS at all — cloud-init no longer clones.

Sharing the base compose verbatim is deliberate: the service's `environment:`/`volumes:`/`ports:` block keeps exactly one definition, so the two modes cannot drift.

Consequences worth calling out:

- **The version-skew bug class becomes structurally impossible** in the default mode — the version *is* the contract.
- **Base-image fingerprints now match by construction.** The control box's `runtime/` tree is literally the same npm tarball the PC installed, rather than a ref that happens to correspond. (`AGENTBOX_RUNTIME_ROOT` still has to be set, and now `AGENTBOX_CLI_RUNTIME_DIR` too — a container has no parent CLI process to inherit it from.)
- `NODE_ENV=production` is load-bearing in the image: without it Next takes the dev path and dies with ``Couldn't find any `pages` or `app` directory``.

`--ref`/`--repo` keep the clone-and-build path (a dev CLI, whose version was never published, falls back to it automatically); `--package <spec>` pins a different npm spec. `deploy.json` now records which build a control box runs — nothing else on the host or the VPS did — and `hub status` shows it.

Also in here, both prompted by the live test:

- `diagnoseUnhealthy` no longer reports a rate-limited certificate as "Caddy can't reach it, so the upstream port is wrong". It greps Caddy's log and names the real cause.
- `--domain <host>` on `hub deploy hetzner` (the deploy already accepted a domain; the CLI never exposed it) — the way out of a Let's Encrypt rate limit, since an sslip.io name is derived from a recyclable IP.

## Verification

Unit: `resolveHubDeploySource`, the asset resolver, cloud-init with/without the clone, and a drift guard that the shipped compose still yields `hubContainerPort === 8787` / `isFullHubCompose`. Full `pnpm build`, `typecheck`, `lint`, and the web build all green. `npm pack --dry-run` confirms `runtime/hub-deploy/**` ships.

Live on Hetzner:

- **Package mode** — image built in ~60s (vs ~160s of pnpm builds), containers up **2m56s** after start. `/opt/agentbox` holds no repo clone, only the six host-shipped files. In-container: version matches the deploying CLI exactly, `gh 2.96.0` present, all four `AGENTBOX_*` pointers resolve under `/opt/agentbox-cli`, every provider's staged-runtime marker resolves (`hetzner/scripts/install-box.sh`, `e2b/scripts/build-template.sh`, `vercel/scripts/provision.sh`, `digitalocean/…`, `_shared/ctl.cjs`, `docker/Dockerfile.box`), and `/healthz` returns `{"ok":true,"ui":true,"cliEntry":true}`.
- **`--ref nightly` regression** — fully green end to end: healthy over public HTTPS, `relay.controlPlaneUrl` wired, and the e2b bake shared with the control box.

One caveat, external to this change: two package-mode runs landed on a **recycled Hetzner IP** that had already hit Let's Encrypt's 5-certs-per-hostname-per-week cap, so their public HTTPS never came up (the hub itself was healthy on `127.0.0.1:8787` throughout). That is what prompted the diagnostic fix and `--domain`. The `--ref` run on a fresh IP proves the shared Caddy path, which package mode does not modify.

---

> [!NOTE]
> **Medium Risk**
> Changes the default production deploy path and VPS provisioning (cloud-init, compose files, env wiring); mistakes could break new control-box installs, though source mode and extensive tests preserve the escape hatch.
> 
> **Overview**
> **Hetzner control-box deploys now install `@madarco/agentbox` from npm (pinned to the deploying CLI’s version) instead of cloning and building the monorepo on the VPS.** Package mode scp’s a small compose stack (`docker-compose.yml` + `docker-compose.package.yml` + `Dockerfile.package`) from `runtime/hub-deploy/` (staged by `stage-runtime.mjs`); cloud-init skips the git clone. `--ref` / `--repo` still use the old clone-and-build path; dev builds (`0.0.0`) auto-fall back to source. `resolveHubDeploySource` centralizes that choice; CLI flags add `--package` and expose `--domain` on `hub deploy hetzner`.
> 
> **Observability and ops:** `deploy.json` records `HubDeploySource`; `hub status` shows the remote **build** line. Failed HTTPS health checks use an on-VPS curl probe (`describeCaddyHop`) so cert/rate-limit issues aren’t misreported as wrong Caddy upstream ports; recovery hints use the correct compose `-f` list per mode.
> 
> **Images:** New `Dockerfile.package` global-installs the CLI and runs the staged hub bundle; source `Dockerfile` gains `AGENTBOX_DOCKER_CONTEXT`. Docs updated for default npm deploy, flags, and troubleshooting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 230d94ed2b8c6df2140e1b1dd2e23f82b3ab2346. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

PR #261, opened 2026-07-26
feat(hub): `hub update` and `hub destroy`, plus the running build in `hub status`

madarco:
## Why

A control box was a one-way trip. Changing its version meant deploying a **whole new VPS**; getting rid of one meant deleting the server by hand in the Hetzner console and then remembering to clear `~/.agentbox/control-plane`, the `agentbox-hub` SSH block, and `relay.controlPlaneUrl`. And `hub status` couldn't say what version was actually running — which matters now that host↔hub version parity is what makes shared bake fingerprints match.

## The enabling refactor

`deployControlPlaneToHetzner` split into **provision** + **`applyControlPlaneConfig`**. Everything from "read the compose" to the healthz poll is exactly what an update has to redo — a newer hub can want new `.env` keys, a different container port, or new compose assets — so `hub update` re-runs the deploy's own configuration half rather than growing a second implementation of it. No behaviour change for deploy.

## `agentbox hub update`

Moves the existing VPS to a new build in place; the data volume (store, logins, custody, box SSH keys) is untouched.

```bash
agentbox hub update                     # match this CLI's version (default)
agentbox hub update --channel nightly   # newest published nightly | stable
agentbox hub update --package 0.28.0    # an exact version
agentbox hub update --ref my-branch     # switch to building from source
```

Defaulting to *this CLI's* version is deliberate — that parity is what keeps bake records shared.

It re-syncs the firewall's SSH source **over the Hetzner API before it connects**. Only `:22` is IP-locked (`:80`/`:443` stay open so boxes reach the hub from anywhere), so a laptop that changed networks can't ssh in at all — and the API path still works when ssh doesn't, so this repairs a box that has already locked you out. Regenerating the env also refreshes `AGENTBOX_HUB_ADMIN_CIDR`, so direct PC→box SSH keeps working after your IP moves.

## `agentbox hub destroy`

Prints exactly what it will delete — VPS, firewall, `~/.agentbox/control-plane`, the `agentbox-hub` alias, `relay.controlPlaneUrl` — and confirms first. `--keep-credentials` keeps `control-plane.env` for a later redeploy.

Two deliberate behaviours:

- **It asks the hub what boxes it still owns and refuses without `--force`.** Cloud boxes outlive the hub and its store is often their only record. An unreachable hub says so rather than blocking, since a broken deploy is exactly what you'd want to delete.
- **It never throws for an already-gone resource**, so the local purge always runs. A half-deleted control box must not leave config pointing at nothing with no command that clears it.

Teardown needs no ssh at all, so it works from a different network than the one that deployed the box. `unset-url --purge` now shares the same purge helper, which also fixes it leaving a stale `Host agentbox-hub` block behind.

## `hub status`

```
hub: remote (reachable)
  url:     https://<ip>.sslip.io
  health:  0 box(es), 0 event(s)
  version: 0.28.0-nightly.202607260716
  channel: nightly
  build:   @madarco/agentbox@0.28.0-nightly.202607260716 (npm)
```

The relay already reported `AGENTBOX_CLI_VERSION` on `/healthz`; both hub images now export it, read from the **installed package's manifest** rather than the requested spec (a spec can be a dist-tag like `nightly`). The live version beats the deploy record — the record only says what was last *deployed*, which is wrong after a failed update and absent for a hub someone else set up. `channel` is `source (<ref>)` for a build-from-source box, since the ref is the distinction that matters there. A drift nudge appears only when there's a **live** reading to compare, so a stale record can't nag forever.

## Verification

Unit: 8 tests for the status precedence (live beats record, fallback, source-ref channel, drift only when live) and 4 for destroy tolerance (already-gone server still deletes the firewall, ordering, never throws, no-op). Dockerfile drift guards assert the version comes from `package.json` and **not** from `AGENTBOX_SPEC`. Full `pnpm build` / `typecheck` / `lint` / suites green, plus the web build.

Live on Hetzner — deploy → status → `update --package <older>` → `update` back → `update --ref` → `destroy` (with and without a registered box). Results in a follow-up comment.

---

> [!NOTE]
> **Medium Risk**
> Changes touch production control-plane provisioning, firewall rules, and destructive teardown; mistakes could orphan cloud boxes or leave broken local config, though destroy gating and non-throwing partial teardown reduce that risk.
> 
> **Overview**
> Closes the Hetzner **control box lifecycle**: **`agentbox hub update`** upgrades the existing VPS in place (data volume kept) via the same configuration path as deploy, and **`agentbox hub destroy`** removes server, firewall, local deploy state, SSH alias, and `relay.controlPlaneUrl`, with guards against orphaning registered boxes unless **`--force`**.
> 
> The enabling change in **`@agentbox/sandbox-hetzner`** extracts **`applyControlPlaneConfig`** from deploy so update re-ships compose, env, Caddy, secrets, and **`docker compose up`**. Updates can **re-sync firewall SSH** to the current egress IP over the API before SSH, and post-health checks use **`docker compose ps`** so a one-off healthy `/healthz` during a crash loop is not treated as success.
> 
> **`hub status`** for remote hubs now shows **version** (live `/healthz` over deploy record), **channel**, **build line**, and a **drift nudge** toward `hub update` only when a live version differs from the CLI. Hub **Dockerfiles** set **`AGENTBOX_CLI_VERSION`** from the installed `package.json`, not the npm spec/dist-tag.
> 
> **`unset-url --purge`** uses **`purgeLocalControlPlaneState`** so stale **`agentbox-hub`** SSH config is cleared. Docs and unit tests cover destroy gating, status precedence, and destroy tolerance for already-deleted cloud resources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ba60531559c9f91fbc372cb6d04dacd8ff66090. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
## Live verification (Hetzner, one VPS)

| Step | Result |
|---|---|
| `hub deploy hetzner` | healthy; `hub status` shows `version: 0.28.0-nightly.202607260716`, `channel: nightly`, no drift nudge |
| `hub update --package <older nightly>` | version really changed **on the same server id** — proves the healthz version is live, not echoed from `deploy.json` |
| `hub update` (no flags) | recovered the box in place; success line reports `running 0.28.0-nightly.202607260716` |
| `hub update --ref nightly` on a package box | refused with an actionable message, box untouched |
| `hub destroy` | prompt listed all five items; server + firewall deleted (firewall retry fired once) |

Post-destroy, all six claims verified independently: server gone, firewall gone, `~/.agentbox/control-plane` gone, `agentbox-hub` SSH block gone, `relay.controlPlaneUrl` unset, and `hub status` correctly falls back to the local hub.

### Two bugs the live run caught (both fixed in 0a87bd4b)

**1. A crash-looping update reported success.** The first downgrade printed *"Control box is healthy"* while the container restarted every 60 seconds. The hub starts listening **before** it starts its create worker, so a build that boots and then throws serves healthz for a moment and dies — and the poll landed in that window. Now `applyControlPlaneConfig` checks `docker compose ps` after the poll (a crash-looping container reports `restarting`, never `running`) and fails with the app logs attached. Re-running the same downgrade afterwards correctly failed.

**2. `hub update --ref` on a package-mode box** died with a bare "not a git repository" — cloud-init never clones one there. It now explains that and points at `hub deploy hetzner --ref`.

### A note on why that downgrade crashes

Not a defect in this PR: `0.28.0-nightly.202607251935` predates the boot-throw fix from the `hub.gitAuth` work, so an older hub cannot start against a `gh`-mode env (`no GitHub App configured`). It is a genuine example of what `hub update` should refuse to call a success — which is exactly what it now does.

### Coverage

The destroy safety gate is split out as the pure `destroyGate` and unit-tested (refuses on a non-empty box list; allows with `--force` while naming the orphans; never blocks on an unreachable hub) rather than spending a live cloud box on it. 121 hetzner + 991 CLI tests green.

madarco:
bugbot run

madarco:
bugbot run

PR #262, opened 2026-07-26
feat(hub): `agentbox hub expose` — make a machine you own the control box

madarco:
## Why

A control box needed a VPS. But plenty of people already have an always-on machine — a Mac mini, a NAS, a desktop — and the hub they run on it is *the same process* a deployed control box runs. `agentbox hub expose` flips it: same hub, now in the deployed profile (password auth, SQLite store, resident create worker), reachable to boxes. No second machine, no bill.

## Provenance

Most of this was built by an agent in a box and checkpointed as `9d126eac` when that box ran out of memory (zombie processes, defunct tmux server). The work was already sound — typecheck, lint and the full suites passed, and `hub update`/`hub destroy` already branched to local flows — so this PR is that checkpoint plus what was missing. It's based on `nightly` after #261.

## What's in it

**`hub expose`** — writes an exposed record to `deploy.json`, restarts the hub in the deployed profile, and points the CLI at it. `ensureHub()` reads that record on every start, so `hub start`, `restart`, autostart and the post-update refresh all reconstruct the mode from disk; nothing has to be re-run.

**Reaching it from cloud boxes** — the one thing a VPS gets for free. `--tunnel cloudflare` (quick tunnel, zero account, ephemeral hostname that AgentBox re-reads on each restart), `--tunnel tailscale` (stable, needs an authed tailnet), or `--public-url` for your own proxy. With no tunnel the hub is advertised at its LAN address and the command **says outright** that cloud boxes can't reach it, rather than letting you discover that at the first create.

**Autostart** — a launchd agent / systemd user unit that just runs `agentbox hub start`. It carries no secrets and no flags, so it can't drift from the expose settings.

**`hub unexpose`** (added here) — the same teardown as `hub destroy`, under the honest name: nothing is destroyed except the exposed mode. `destroy`'s description no longer claims a VPS unconditionally.

**Docs** (added here) — none existed. A "two ways to get one" comparison table up front, then a section covering the tunnels, binding, the autostart login-vs-boot caveat per OS, and teardown.

## A gate bug the live test found

The first `hub unexpose` **refused**, reporting one registered box — the local `docker` box the work came from. But a docker box runs on this machine against the host's own `.git` and is never handed to the control box, so the hub going away cannot strand it. `destroyGate` now counts only boxes that could actually be orphaned; `remote-docker` is deliberately **not** exempt, since that one lives on another machine. The gate also returns the orphans it counted, so both call sites list what's actually at risk instead of everything registered.

## Verification

Live, end to end on a real machine — expose (loopback, `--no-autostart`) → `hub status` shows `mode: exposed (control box, worker on)` and `/healthz` reports `profile: hetzner, worker: true` → the gate refuses, then allows once the docker box is correctly ignored → `unexpose` → **every value back to the pre-test snapshot** (no control-plane dir, `relay.controlPlaneUrl` unset, no launchd unit) and the hub back on `profile: localhost`.

One trap worth recording: `hub status` initially showed no exposed block because the *staged* hub bundle was stale — it's rebuilt by `build:standalone`, not by the CLI build. The process env was correct all along. Exactly the daemon gotcha CLAUDE.md warns about.

Gates: `pnpm build`, `typecheck` (31), `lint` (17), the web build, and 1,948 tests across cli / sandbox-core / relay / sandbox-docker all green.

---

> [!NOTE]
> **Medium Risk**
> Changes hub auth profile, public reachability (tunnels), and control-plane teardown gates—areas that affect whether cloud boxes can register and whether teardown is safe—but behavior is covered by new unit tests and documented failure modes.
> 
> **Overview**
> **`agentbox hub expose`** turns the existing local hub into the deployed control-box profile (password auth, SQLite store, resident worker) without provisioning a VPS. It writes a `provider: 'local'` deploy record, sets `relay.controlPlaneUrl` to the box-facing URL, restarts the hub via `ensureHub()`, and optionally installs launchd/systemd autostart. **`hub setup --deploy local`** and interactive setup now default toward this path.
> 
> **Reachability for cloud boxes** is opt-in: **`--tunnel cloudflare`** (quick or named token) or **`tailscale`**, plus **`--public-url`** when you terminate TLS yourself. Tunnel lifecycle is tied to **`hub start`/`stop`/`restart`** (restore/stop helpers), with fixes so tunnels actually start when `--public-url` is set, named tunnels require `--public-url`, bad flag combos fail before tearing down a working tunnel, and ephemeral Cloudflare hostnames rotate into config + hub env.
> 
> **`hub unexpose`** is the local teardown alias (same as **`hub destroy`** for `provider: 'local'`): tunnel, autostart, exposed mode, and config — shared `~/.agentbox` kept. **`hub update`** on a local box restarts the exposed hub in place (CLI upgrade is `self-update`).
> 
> The CLI now prefers **loopback** (`127.0.0.1:<port>`) for custody/API/target when this machine is exposed, so tray/CLI don't use the LAN/tunnel URL or the wrong token after `unset-url`. **`hub status`** surfaces exposed config even when the hub process is stopped.
> 
> **`destroyGate`** only counts boxes that would be orphaned (**excludes `provider: docker`**; **`remote-docker` still blocks**) and returns the orphan list for prompts.
> 
> Docs: public **deployed-hub** / **cli** pages for expose vs VPS; internal doc map trimmed (e.g. backlog links consolidated).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd6717778d8c423bd9a6c5be9db2b4ff3fc80ec8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
## Re-verified live after the review fixes

Full round trip against the final code, seeded with a **stale `AGENTBOX_TUNNEL_TOKEN`** to exercise the state bug:

| Check | Result |
|---|---|
| `hub expose --bind 127.0.0.1 --no-autostart` | flips; warns that cloud boxes can't reach a loopback hub |
| stale tunnel token | **cleared** by the expose (was: silently reused → named tunnel on an unknown hostname) |
| `hub status` running | `mode: exposed (control box, worker on)`; healthz reports `profile: hetzner, worker: true` |
| `hub target` | `hub: remote` + the API key (was: local mode + a token the exposed hub rejects) |
| `hub status` **stopped** | still reports `mode: exposed (configured; the hub is not running)` |
| `hub unexpose` | stops hub + tunnel, purges state |
| after | control-plane dir absent, `relay.controlPlaneUrl` unset, no launchd unit — identical to the pre-test snapshot; hub back on `profile: localhost` |

Bugbot is clean on `e68cba7`; CI green. 1,018 CLI tests + the other suites pass.

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #265, opened 2026-07-27
feat(hub): Custody and System/Build pages in the hub Web UI

madarco:
## WS3 — hub Web UI: Custody + System/Build pages

Adds the two read-only pages from the remote-hub improvements backlog (items 3 & 4), each a pure REST client over a new `/api/v1` route.

### Item 3 — Custody page (`/custody`, `GET /api/v1/custody`)
Lists what the control box holds so a box created from either side is usable from both — agent credentials, project seeds, provider bake records, and per-box SSH keys — grouped by scope (agents / projects / prepared / boxes) with each entry's path, short sha256, size, and last-write time. **Value bytes are never returned** — same contract as `agentbox hub custody list`. `enabled: false` on a hub with no custody (localhost) renders an explanatory empty state. `?prefix=` scopes the listing; a malformed prefix is rejected `400`.

### Item 4 — System & Build page (`/system`, `GET /api/v1/system`)
Answers "what is running here, and do I need to re-bake?": hub version + channel + build source, the deploy record (when the machine is an exposed/deployed control box), each base provider's baked fingerprint + freshness (a `stale` row is exactly when to re-run `agentbox prepare --provider <id>`), and the box-image build-context manifest (skills / agents / scripts baked in). Reuses the existing freshness logic (`backend.providersWithFreshness`) and `@agentbox/sandbox-core` prepared-state rather than reimplementing it. Freshness is authoritative: when the backend reports a base as `unprepared` (image/snapshot actually absent), it renders "not baked" even if a stale on-disk record lingers.

### Shape / conventions
- Pure REST client per `feedback-hub-web-pure-rest-client`: client-side fetch, no server actions. Both routes registered in `openapi.ts`. Auth copied from existing routes — Bearer for `/api/v1`, session cookie for the browser.
- Custody store reached via `globalThis` (set by the custom server), so `@agentbox/relay` stays out of Next's bundle.
- Pure presentation helpers in `lib/custody-view.ts` + `lib/system-info.ts` with vitest unit tests (19 tests). Sidebar entries + `key`/`server` icons. No new runtime dependencies (adds `vitest` as a hub devDep).
- Public docs updated in the same change: `api.mdx` (both endpoints) and `deployed-hub.mdx` (both pages).

### How it was smoke tested
Built the standalone bundle and launched it directly with a control-box env (`AGENTBOX_HUB_PROFILE=hetzner`, an admin token so custody is wired), seeding fake custody entries across all four scopes. Verified against the running hub:
- `GET /api/v1/custody` returns metadata only across all scopes; `?prefix=agents` filters correctly; a traversal-y prefix (`../etc`) → `400`; unauth → `401`.
- `GET /api/v1/system` returns real data — hub version/channel (nightly detected from the `-nightly.` marker), per-provider prepared fingerprints + freshness (e2b rendered `stale` with its real `baseStaleReason`; an absent docker base correctly rendered not-baked after the freshness fix), and the build-context manifest.
- Both endpoints accept **both** a `Bearer` key and a better-auth session cookie.
- Signed in via better-auth and loaded `/custody` and `/system` in the browser path — both serve `200` with the expected headings and sidebar entries, no error boundary.

Verification: `pnpm --filter @agentbox/hub typecheck`, `lint`, `test` (19 passing), and `pnpm turbo run build --concurrency=1 --filter=@agentbox/hub` all green.

https://claude.ai/code/session_01Td3inHnLSND7WNCwQ5wHLy

---

> [!NOTE]
> **Low Risk**
> Read-only metadata endpoints and UI; custody never returns secret bytes and deploy fields are explicitly whitelisted, with no changes to auth or write paths.
> 
> **Overview**
> Adds **Custody** (`/custody`) and **System & Build** (`/system`) to the hub dashboard, each a client-side fetch against new read-only API routes—no server actions.
> 
> **`GET /api/v1/custody`** returns custody manifest **metadata only** (path, size, sha256, mode, `updatedAt`); secret values are never returned. It reads the store via `globalThis.__AGENTBOX_HUB_CUSTODY` (set in `server.ts` when an admin token enables custody) and responds with `enabled: false` on hubs without custody. Optional `?prefix=` is validated and rejects malformed scopes with `400`.
> 
> **`GET /api/v1/system`** answers version/channel/build, a **whitelisted** deploy record slice, per–base-provider bake fingerprints plus freshness from `backend.providersWithFreshness` (with `unprepared` overriding stale on-disk records), and box-image build-context keys from `sandbox-core`.
> 
> OpenAPI gains **Custody** and **System** paths/schemas; sidebar adds both pages plus a **key** icon. Pure helpers live in `lib/custody-view.ts` and `lib/system-info.ts` with **vitest** (hub `test` script). Public docs in `api.mdx` and `deployed-hub.mdx` describe the endpoints and pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e99c258bef51f2eb433fbdbbdc83b421f153f22. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
> Bugbot: `baseStatus === "fresh" || === "stale"` maps "unknown" to baked=false, dropping the fingerprint for a base that is actually prepared.

Confirmed and fixed in 3635ff6d.

You are right about the semantics: per `baseFreshnessFromFingerprints` in `packages/sandbox-cloud/src/checkpoint.ts`, `unprepared` means there is **no** stored fingerprint (genuinely not baked), whereas `unknown` means a stored fingerprint **exists** and only the live one could not be computed — so the base *is* baked, freshness is merely unverifiable. (`classifyDockerBaseFreshness` agrees: `unknown` requires `imagePresent`.) Reporting "not baked" for `unknown` was exactly the wrong answer for the question the System page exists to answer.

The fix:
- Extracted `isBaked(baseStatus, hasRecord)` — only `unprepared` counts as not-baked (`baked = baseStatus !== "unprepared"` when freshness is present); falls back to the on-disk record when freshness is absent (the plane read path). The route now keeps the prepared metadata (fingerprint / cliVersion / bakedAt) for `unknown`.
- `bakeVerdict` and the status badge render `unknown` honestly as **baked-but-freshness-unverifiable** ("Baked — freshness could not be verified here." / a neutral `baked · unverified` badge, no green run pill) rather than as either fresh or stale.
- Added a unit test covering all four `baseStatus` values for `isBaked` (plus the no-freshness fallback) and a `bakeVerdict` case for `unknown`, so this cannot regress.

Verified: `pnpm typecheck`, `lint`, `test` (22 passing), and a filtered serial build — all green.

PR #276, opened 2026-07-28
feat(hub): stream a hub create's progress back to the PC

madarco:
## Why

`agentbox e2b claude` (and `create --via-hub`, `codex`, `opencode`) routed to a control box showed **one spinner line stuck on `remote hub: running`** for the 5-10 minutes a cloud create takes. The job row (`CreateJobRow`) carries a status string and a terminal result — nothing else — so the poll loop could only ever report two transitions.

The progress already exists: the hub worker appends every step (lease, clone, seed overlay, the provider's own create output) to `~/.agentbox/logs/queue-<jobId>.log` **on the control box**, which the hub web UI streams. The PC never fetched it, and `~/.agentbox/logs/<command>.log` was effectively empty for hub runs.

## What

- **relay** — `GET /remote/boxes/:id/logs?offset=<bytes>` on the same admin-bearer-gated dispatcher the create already uses. Returns `{ lines, offset }`; the offset only ever advances past complete lines, so a read landing mid-line never truncates or drops one. The dispatcher stays fs-free: the daemon injects `readJobLog` (`job-log-tail.ts`), the serverless plane doesn't and answers 501. Job ids are validated before any path join.
- **CLI** — `pollHubJob` gains `onLog`, draining the tail between status checks (2s tick when tailing) and once more after the terminal status. A hub without the route (404/405/501) disables tailing silently and the create proceeds status-only.
- **rendering** — `withHubJobLine` now uses `makeProgressReporter`, the same helper the local create path uses: collapsed self-updating line by default, full streamed lines under `-v`. Every line is teed to `~/.agentbox/logs/<command>.log`; the background `-i` hub path passed no `onLog` at all before, so that log was empty.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm lint`, full test suite green. New unit tests: route (offset, 501 without the reader, 401 before reading), `readCreateJobLog` (partial line held back, truncation restart, traversal refused), and the CLI poll (offset progression, graceful fallback on an old hub).
- **Live wire**: a real relay daemon + the real CLI `pollHubJob` against a real job log written with deliberate partial lines — all lines delivered in order, none duplicated or truncated.
- **Live CLI**: `create --provider e2b --via-hub` against a stand-in control box (relay daemon + a worker that writes the same per-job log). The spinner tracks the worker's lines through to `box ready: e2b-7a1f`; all 7 lines land in `~/.agentbox/logs/create.log`; `-v` streams them unclamped.

Not run: a real cloud create through a deployed VPS hub — the worker-side log writing is unchanged code (the hub UI consumes the same file).

## Docs

`apps/web/content/docs/deployed-hub.mdx` (what you see while waiting, `-v`, the local transcript) and `docs/architecture.md` (the new route).

https://claude.ai/code/session_01QUnCvB6dh6JWknXFi5pSzk

---

> [!NOTE]
> **Low Risk**
> Observability and UX around an existing admin-gated create queue; log read failures are non-fatal and job IDs are validated before any path join.
> 
> **Overview**
> Hub-routed creates (`create --via-hub`, foreground/background `claude`/`codex`/`opencode`) no longer sit on a static **`remote hub: running`** line for minutes. The relay adds **`GET /remote/boxes/:id/logs?offset=<bytes>`** (admin-bearer, same surface as create) so the CLI can tail the worker’s per-step log; the daemon wires **`readCreateJobLog`** from **`job-log-tail.ts`**, while planes without a log file answer **501** and polling falls back to status-only.
> 
> **`pollHubJob`** gains **`onLog`**: it drains new lines between status ticks (2s when tailing), retries transient log errors, stops tailing only when the hub permanently lacks logs, and does a final drain when the job finishes. **`hubLogSink`** feeds the live status line (ISO timestamps stripped) and **`~/.agentbox/logs/<command>.log`** (verbatim). **`withHubJobLine`** now uses **`makeProgressReporter`** with **`-v`** for full streamed output, matching local creates; agent hub paths that omitted **`onLog`** now tee worker output into the command log.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58d401f9b6edcb341530869354b1970bec63efb4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

PR #279, opened 2026-07-29
feat(cli): make the local hub an /api/v1 target (thin CLI Step 0)

madarco:
## What

Step 0 of `docs/hub-api-single-path-plan.md` — **make the local hub an `/api/v1` target**. This is the blocking prerequisite for routing every CLI box operation through `/api/v1` in both local and remote modes.

Before this, `resolveHubApiTarget` hard-required `relay.controlPlaneUrl` **and** `AGENTBOX_HUB_API_KEY`, so every `/api/v1` client path in the CLI was remote-only by construction — a local hub was never a server the CLI talked to.

## Why

"Thin" was a property of the remote hub only. Making the local hub an API target is what lets steps 1–14 be a base-URL swap rather than an `if (remote)` fork in every command.

## Changes

- **`resolveHubApiTarget` now delegates to `resolveHubTarget`** (`commands/hub.ts`), which already resolves all three shapes — a configured control box, the `hub expose` loopback, and the local hub (`127.0.0.1:<port>` + `~/.agentbox/hub/token`). The local hub's `proxy.ts` already accepts that token as `Authorization: Bearer`, so `HubApiClient` needs no change. It returns a URL-swappable `{ url, apiKey }` for both modes.
- **Local-hub autostart.** When the target resolves local and the hub isn't running, it is started via `ensureHub` (one spinner, one clear failure line). Skipped under `quiet` so a best-effort probe never spawns a daemon.
- **`withHubClient()`** (new `control-plane/with-hub.ts`) — the single entry point later steps wrap an op in. It resolves the client, runs the `GET /api/v1/health` **version gate** (`apiVersion` ∈ `SUPPORTED_HUB_API_VERSIONS = ['v1']`, refusing an unsupported hub with an upgrade hint), then runs the op and maps `HubApiError` codes → stable CLI exit codes + actionable messages (`not_found`=2, `unauthorized`=3, `invalid_request`=4, `conflict`=5, `backend_unavailable`=6, else 1). The `hub boxes` group is converted to it as the canonical example.
- **Module cycle.** `hub.ts` and `control-plane.ts` sit in a cycle (`hub.ts` consumes `controlPlaneSubcommands` at load), so the `resolveHubTarget` / `resolveHubApiTarget` edges are imported **lazily** — matching the existing pattern in this file.
- Docs: `cli.mdx` `hub boxes` note (now targets the local hub too, autostarting it); Step 0 ticked in the plan doc with notes for later steps.

## How verified (end-to-end, ground truth)

- **Local hub, no control box** — the acceptance criterion (errored before): `agentbox hub boxes list` autostarts the local hub, passes the version gate, and lists boxes. Exercised **empty** (`No boxes on the hub.`, exit 0) and **non-empty** (a real docker box created from `examples/express-ready`). Confirmed against ground truth: `GET /api/v1/health` reports `apiVersion: "v1"`; `GET /api/v1/boxes` (local token as Bearer) returns the same boxes.
- **Lifecycle + error mapping** — `hub boxes pause <id>` (exit 0), `hub boxes pause <nonexistent>` → `not_found` (exit 1).
- **Remote-shaped path** via `agentbox hub expose --bind 127.0.0.1 --no-autostart`: `hub target` reports `mode: remote` with the API key; `hub boxes list` lists via the key-authenticated path. Ground truth: `/api/v1/boxes` returns **200 with the key, 401 without**. Teardown (`hub unexpose`) verified clean.
- `pnpm --filter @madarco/agentbox typecheck` clean; full CLI suite `1095 passed | 1 skipped` (17 new/updated assertions across `with-hub.test.ts` + `hub-api-client.test.ts` covering the three target shapes, version gate, error mapping, and the health probe); eslint clean on touched files.

## Notes for later steps

Converting a command is `await withHubClient(opts, async (client) => { … })`; a command that special-cases a code (e.g. `not_found` → info) catches it **inside** the callback before it reaches the mapper. Keep the `resolveHubTarget` / `resolveHubApiTarget` imports lazy.

https://claude.ai/code/session_01XyCfcZoTxhKUgf5riou9TP

---

> [!NOTE]
> **Medium Risk**
> Touches hub target resolution and can autostart Docker-backed local hubs on CLI use; behavior change for users without a configured control box, but scoped to `/api/v1` entry points and covered by tests.
> 
> **Overview**
> **Step 0** of the hub `/api/v1` consolidation: the CLI is no longer remote-only for `HubApiClient` — local and control-box targets share one URL + Bearer shape.
> 
> **`resolveHubApiTarget`** now follows **`resolveHubTarget`** (local token vs `AGENTBOX_HUB_API_KEY`) via **`hubApiTargetFrom`**, instead of requiring `relay.controlPlaneUrl` and a control-plane env key. For a **local** hub that is down, it can **autostart** the full hub (`ensureHub` + `rehydrateFromState`), gated on `running && ui` so a bare relay on the port is not mistaken for the API. **`quiet`** skips autostart.
> 
> New **`withHubClient`** resolves the client, probes **`GET /api/v1/health`**, refuses unsupported **`apiVersion`** values, and maps **`HubApiError`** codes to stable exit codes and hints. **`hub boxes`** (list, lifecycle, rm) is refactored to use it as the pattern for later commands.
> 
> **`HubApiClient.health()`**, **`SUPPORTED_HUB_API_VERSIONS`**, and unit tests cover health, target shapes, version gate, and exit-code mapping. Docs note local-hub autostart for `hub boxes`; the plan marks Step 0 done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36ce91a654d90d0b4f372c485e68933f89f1ab69. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/control-plane.ts
+++ b/apps/cli/src/commands/control-plane.ts
@@ -109,6 +109,7 @@ import {
   HubApiError,
   type HubLifecycleAction,
 } from '../control-plane/hub-api-client.js';
+import { hubApiTargetFrom, withHubClient } from '../control-plane/with-hub.js';
 import { loadControlPlaneEnv } from '../control-plane/env-file.js';
 import { getHubJob, listHubJobs } from '../control-plane/hub-enqueue.js';
 import type { CreateJobRow } from '@agentbox/relay/control-plane';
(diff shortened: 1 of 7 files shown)


madarco:
bugbot run

madarco:
bugbot run

PR #284, opened 2026-07-29
feat(cli): route box resolution through the hub /api/v1 (thin CLI Step 4)

madarco:
## What & why

Step 4 of `docs/hub-api-single-path-plan.md` (one path: every CLI box op goes through `/api/v1`). The IO plane stays direct, so the CLI still materializes a local `BoxRecord` to resolve a box — but that materialization (adoption) previously reached the control box over the **internal** `/admin/store` wire (`ControlPlaneAdminClient.listBoxes` + `matchRegistration`). This re-sources adoption onto the **public** `/api/v1`, the same surface the tray/web use, so the internal wire is no longer a client dependency.

## What changed

**Server (`apps/hub`)**
- New `GET /api/v1/boxes?ref=<id|name|index>` — server-side resolution via the pure `resolveBoxRefView` (`apps/hub/lib/boxes/resolve.ts`): `findBox` precedence on the topology-agnostic `Box` view (exact id → unique id prefix → name → displayName → sandbox id / `cloud:<id>`), plus the numeric project-index arm when `?project=` is supplied. Returns the **match set** as `{ boxes }` — 0 (none) / 1 (unique) / >1 (ambiguous prefix) — so ambiguity is expressed rather than arbitrarily narrowed. Operates on the full box list (incl. registered-only rows), so hub-created boxes are found by name.

**Client (`apps/cli`)**
- `HubApiClient.resolveBox(ref, project?)`.
- **Resolution stays local-first.** `box-ref.ts` is unchanged: `resolveBoxOrExit` still runs the local `resolveBoxRef` (index/autopick/ambiguous, offline) and only on a local miss calls `tryAutoAdopt` — which is what got re-sourced (from `/admin/store` onto `resolveBox`). Routing every resolve through the hub would add latency + a network dep to every box-arg command; keeping local-first avoids that.
- The fallback engages only for a genuinely remote/exposed hub (`resolveHubTarget().mode === 'remote'`, reusing Step 1's exposed-loopback-first ladder), never a plain local hub — so a typo on a laptop with no control box never round-trips or auto-starts a daemon.
- `adoptHubBox` / `pullBoxSshKeys` now take an already-resolved `HubApiBox` and materialize via the shared `registrationToBoxRecord` (`hubBoxToRegistration` bridges the payload → `BoxRegistration`). The per-box SSH-key pull stays on custody (`/admin` — Step 10 moves it); `custody` is optional so a keyless thin client still adopts the record (SSH providers flagged `sshKeysMissing`).
- `hub adopt` / `hub pull` resolve through `/api/v1` too. `control-plane/match-ref.ts` deleted (matching is server-side). `_cloud-agent-via-hub.ts` + `recover.ts` updated to the new signatures.

**Registration-side `originUrl` (Step 4's explicit second half)**
- Capture `originUrl` at registration for docker (`create.ts`) and the classic-cloud create/resume paths (`cloud-provider.ts`) — the registration side of Step 3's cross-machine project scoping (the control-plane path already sent it). Kept to the `originUrl:` argument alone so Step 8 (which also edits these create paths) rebases cleanly.

**Docs**: `api.mdx` + `openapi.ts` document the `?ref`/`?project` params; plan doc ticked + a "Notes for later steps" block added.

## How verified

- `pnpm typecheck` (cli, hub, relay, sandbox-docker, sandbox-cloud) and `pnpm test` green. New unit tests: `apps/hub/test/resolve.test.ts` (resolver: all match kinds incl. ambiguous + index), rewritten `hub-adopt.test.ts` / `hub-pull.test.ts` (materialization from a resolved `HubApiBox`, every provider shape, project linkage, idempotency, no-custody).
- **End-to-end** against a remote-shaped loopback hub (`relay.controlPlaneUrl` → `127.0.0.1`, `AGENTBOX_HUB_API_KEY`): a registered-only **e2b** box resolved through the route (`GET /api/v1/boxes?ref=calm-otter`) and `agentbox url <name>` **adopted it with no explicit `hub adopt`** — the materialized `BoxRecord` landed in `state.json` with `control-plane` topology and was **project-linked via its `originUrl`** to a local clone.
- Note: hetzner/e2b via a real control box couldn't be run here (the local-expose setup needs a GitHub token this env can't extract, and the memory-heavy standalone hub intermittently stalls its `/api/v1/boxes` route — which calls `docker ps` — under this VM's memory pressure while `/health` stays green). The adoption path is proven above with an e2b-shaped box; the SSH-provider (hetzner) difference is only the custody key pull, which is unit-covered. The numeric-index *local* path is unchanged code (`box-ref.ts` untouched) and unit-covered on the server side.

https://claude.ai/code/session_01J9DXGPJnagS2EUwv1ZdaRF

---

> [!NOTE]
> **Medium Risk**
> Changes how every box-arg command resolves hub-only boxes (local-first, then bounded auto-adopt) and how SSH keys are keyed; mistakes could target the wrong box or adopt without keys, but behavior is heavily tested and gated on remote hub + API key.
> 
> **Overview**
> **Hub box resolution moves onto the public `/api/v1` surface** instead of the internal `/admin/store` registry. The hub adds `GET /api/v1/boxes?ref=` (optional `?project=`) with `resolveBoxRefView`, returning a **match set** `{ boxes }` that mirrors CLI `findBox` semantics (including ambiguous id prefixes and numeric project index). Postgres-backed box rows now expose `name` for name-based refs.
> 
> **CLI adoption and pull** resolve refs via `HubApiClient.resolveBox`, then materialize from an already-resolved `HubApiBox` (`hubBoxToRegistration` → `registrationToBoxRecord`). `tryAutoAdopt` only runs against a **remote** hub (`resolveHubTarget`), passes `projectRoot` for index refs, and treats ambiguous hub matches like local resolution. `hub adopt` / `hub pull` share `resolveHubBox`; custody SSH key download stays on `/admin` and is optional for API-key-only clients (`sshKeysMissing`). **`match-ref.ts` is removed**; `recover` downloads keys from the local record without re-resolving.
> 
> **Registration** now records **`originUrl`** at docker and classic-cloud create/resume so thin clients can link adopted boxes to local clones. Docs/OpenAPI and unit tests cover the resolver and updated adopt/pull behavior; the via-hub create path still adopts via admin `listBoxes` until a later step.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3e1cb009434053f0f20bc483bccbf97f3126620. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
Fixed in b60ed20f. The `--via-hub` create flow now materializes the just-created box by resolving it from the control box's store by exact id using the **admin token it already holds** (the same credential it used to enqueue, poll, and seed custody), instead of depending on `AGENTBOX_HUB_API_KEY`. So a successful create always adopts, even on an admin-only machine — restoring the pre-Step-4 behavior. A small transitional reg→HubApiBox adapter feeds the unchanged `adoptHubBox`; Step 8 removes it when create moves fully onto /api/v1.

bugbot run

madarco:
bugbot run

madarco:
Fixed in 04429a54. The Postgres-backed `readState` source now sets `name: r.name` on its box view (the in-process host source already did), so `GET /api/v1/boxes?ref=<name>` and the CLI's `resolveBox` match by name on that topology too. I deliberately did not loosen `findBoxView` to fall back to `task`, since `task` is the display label (a session title when present) and would make refs match session titles.

madarco:
bugbot run

madarco:
Fixed in c3e1cb00. Two parts:
- `tryAutoAdopt` now threads the cwd project root through to `resolveBox`, so a numeric ref keeps project-index semantics server-side.
- `resolveBoxRefView` resolves a numeric ref index-first (when a project is given), then matches **only exactly** (id/name/displayName/sandbox id) — never by id prefix. This kills the wrong-box/ambiguous-prefix risk for refs like `2` while preserving `attach <numeric-sandbox-id>` (hetzner/DO ids are typed in full, so they match exactly). `resolveBoxOrExit` only reaches the hub after its local index lookup already missed, so this never shadows a real local index.

madarco:
bugbot run

PR #285, opened 2026-07-30
feat(cli): route services, rename, url/screen through the hub /api/v1 (thin CLI Step 7)

madarco:
## Step 7 — services, rename, url/screen onto `/api/v1`

Part of the [thin-CLI consolidation](docs/hub-api-single-path-plan.md): the CLI calls the hub's public `/api/v1` for every box op, against a local hub and a remote control box alike. Step 7 converts the last fleet-ish box ops whose routes already existed but were unused by the CLI.

### What changed
- **`services` / `services restart`** → `GET|POST /api/v1/boxes/:id/services*` via `withHubClient`. The route runs the box's `provider.exec` **hub-side** (where the credentials live for a remote control box) using the same shared `boxServicesStatusRaw` / `boxRestartService` core the CLI called inline — a caller-swap, not a rewrite. A **paused/stopped** box now falls back to the hub's persisted snapshot, so `services` finally **agrees with `status`** (the old inline path errored *"could not reach the supervisor"*).
- **`status --set-name` / `--clear-name`** → `POST /api/v1/boxes/:id/rename` (the same `setBoxDisplayName`, now hub-owned).
- **`url` / `screen`** read the box's URL off the enriched Box payload (`webUrl` / `vncUrl`, Step 3) for **docker** instead of probing the provider, starting the box first **only when it isn't running** — with the *"started it"* notice on **stderr** so `--print` stays pipeable. **Cloud keeps the provider `resolveUrl` path**: a cloud box's payload URL is the non-signed `previewUrl` (a header-token URL for Daytona a browser can't open from a click), while `resolveUrl` mints a browser-safe **signed** URL. `--loopback` / `--ttl` also keep the provider path (the payload can't express them), and `screen` keeps its in-box browser prep (box IO).
- Three new `HubApiClient` methods (`getServices`, `restartService`, `rename`) mirror the routes; unit-tested. The shared core in `packages/sandbox-core/src/box-git.ts` is untouched.

### Scope decisions (see the plan's Step-7 Notes)
- **Branches: nothing added.** `/api/v1/boxes/:id/branches` + `/projects/:id/branches` are **web-UI-only** consumers — the CLI has no branch picker (`--from-branch` is a flag). Adding client methods with no caller would be dead code.
- **Flagged for Step 8:** `--from-branch` is host-validated in `lib/from-branch.ts`, which is redundant + incorrect under a remote hub (the PC may have no repo copy; `CreateBoxInput.fromBranch` is already backend-validated). Left for Step 8, which owns `create`.

### Why cloud stays on the provider path
The Step-3 caveat about stale/absent preview entries is real: the payload can't carry a browser-openable signed URL for signed-URL backends, and that live refresh belongs server-side (out of Step 7's file set). Gating the payload path to docker avoids a Daytona regression.

### Verified end-to-end (ground truth, both modes)
Docker box with a declared `expose:` service, tested against a **local hub** and a **`hub expose`d** remote-shaped loopback hub (`AGENTBOX_HUB_API_KEY`):
- `services` agrees with `status` **running** (source `live`) and **paused** (source `persisted`, snapshot pushed via `POST /events`).
- `services restart <svc>` / all / `<unknown>` → exit 5.
- `url --print` → payload URL, **`curl` HTTP 200**; `--loopback` → provider loopback URL.
- `screen --print` → payload noVNC URL + in-box browser prep.
- `--set-name` / `--clear-name` rename and the new label resolves by name.
- **Cloud (daytona container-class box):** `url --print` / `screen --print` return signed HTTPS URLs via the (unchanged) provider path; `url` **`curl` HTTP 200**.

`pnpm typecheck` + `pnpm lint` green; new client-method unit tests pass. Boxes destroyed, no orphan daytona sandbox (API list: 0). Unrelated pre-existing `package.json` tray-rename change left out of this PR.

https://claude.ai/code/session_012vDcaafb3NEW47K2Y28XJd

---

> [!NOTE]
> **Medium Risk**
> Touches user-facing URL/open and service-restart flows and changes `services --json` shape; behavior is intentionally split between hub payload (docker) and provider (cloud/flags), so regressions would show up as wrong URLs or stale endpoints.
> 
> **Overview**
> **Thin-CLI Step 7** moves more box operations onto the hub’s public `/api/v1` via `withHubClient`, so local and remote control-box hubs behave the same.
> 
> **`agentbox services`** no longer calls `provider.exec` from the laptop. List and restart use `GET|POST /boxes/:id/services*`; paused/stopped boxes can show a **persisted** snapshot (aligned with `status`) instead of failing when the supervisor is unreachable. `--json` now returns `ServicesResult` with a `source` field.
> 
> **`agentbox status --set-name` / `--clear-name`** posts to `POST /boxes/:id/rename` instead of calling `setBoxDisplayName` inline.
> 
> **`agentbox url` and `agentbox screen`** (docker, default flags) take `webUrl` / `vncUrl` from `getBox` on the enriched payload, start the box through the hub when not running (notice on **stderr**), and still run in-box browser prep locally. **`--loopback`**, **`--ttl`**, and **cloud** boxes keep the provider `resolveViaProvider` path (signed URLs and options the payload can’t express).
> 
> `HubApiClient` adds **`getServices`**, **`restartService`**, and **`rename`** with unit tests; docs note hub-based URL resolution and auto-start behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75b1d93fa4beffc3789145b1c4e21f9edcfdc6b1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #293, opened 2026-07-30
feat(cli): retire the second implementation + internal client wire (thin CLI Step 11)

madarco:
## Step 11 — retire the second implementation and the internal client wire

The consolidation payoff of the `/api/v1` single-path plan (`docs/hub-api-single-path-plan.md`). After Steps 0–10 converted every CLI box/fleet operation onto the hub's `/api/v1`, this step **deletes what those conversions replaced** and proves the CLI no longer holds a client for the internal relay wire.

### What changed

**Deleted (the internal box/fleet client wire + its now-dead consumers):**
- `control-plane/admin-client.ts` — `ControlPlaneAdminClient` (`/admin/store`, `/admin/prompts`, `DELETE /remote/boxes`).
- `control-plane/reap.ts` — its only consumer (`dashboard.ts`) now destroys cloud boxes through `/api/v1` (`client.destroy`, which does provider teardown **and** store/custody reap server-side).
- `control-plane/hub-list.ts`'s **legacy `/admin/store` half** (`fetchHubListing` + its cache). The `/api/v1` half (`fetchBoxListing`/`cacheAge`/`hostReachable`) stays — `list.ts`/`recover.ts` use it.
- `control-plane/{hub-merge,list-merged}.ts` — replaced by the re-sourced, dashboard-local `dashboard/box-list.ts` (`listDashboardBoxes`/`mergeApiBoxes`), sourced from the **same `/api/v1/boxes` wire `ls` uses** instead of the `/admin/store` registration listing. The merge stays because the dashboard TUI needs local rich fields (`claudeQuestion`, live shell sessions, endpoints) that `HubApiBox` doesn't carry, merged with hub-only rows.
- dead `lib/wait/events.ts` (`waitForEvent`/`/admin/events`, no callers) and obsolete tests.

**Added:**
- `test/no-internal-wire-client.test.ts` — a guard that scans `apps/cli/src` (with a real comment-stripping scanner so `/admin/` in comments is ignored but path literals are caught) and fails on any `/admin/`/`/remote/` path outside a small, justified **(file → path-prefix)** allowlist.
- `dashboard/box-list.ts` + its ported merge/listing unit tests.

**Swept:** `hub-backend.ts`'s "mirrors the CLI's old X" second-implementation comments; fixed doc comments referencing deleted symbols.

### The known tension (Step 10 custody fallback) — resolved by allowlisting

Step 10 kept one CLI `/admin/custody` fallback for byte-reads when only the relay **admin token** (no `/api/v1` API key) is available — without it, per-box SSH keys aren't pulled and `attach`/`cp` break. I **allowlisted** it rather than deleting it. The alternative ("make the `/api/v1` byte-read accept the admin token alone") is infeasible: `/api/v1/custody` is gated by the **hub API key** at the proxy first, so a token-only machine can't reach it at all — which is exactly why the fallback speaks `/admin/custody`. The hub-side byte-read stays fail-closed + loopback-peer-gated (untouched). The allowlist is per-`(file, path-prefix)`, so a *different* `/admin/...` in that file still fails.

### Deliberate deferral — `route-create.ts` kept

The three agent launchers (`claude`/`codex`/`opencode`) still run an **inline local `createBox`** for their **foreground** create path, routed by `route-create.ts`. `route-create.ts` holds **no `/admin`/`/remote` client call**, so the guard test — the real acceptance — passes with it retained. Finishing it is coupled to moving the **create+attach IO plane** behind the hub, which the plan explicitly puts out of scope; Step 8 deferred the launcher conversion for that reason. This is documented as the one remaining inline create path, with an explicit Step 14 qualification so the public docs don't overstate the consolidation. `commands/create.ts` itself already goes through `POST /api/v1/boxes` and does not import `route-create.ts`.

### How it was verified (ground truth, both modes)

- `pnpm build` + `pnpm typecheck` + `pnpm lint` green; **CLI 1115 + hub 120 tests pass**.
- Guard test **genuinely fails** on a temporarily-reintroduced `/admin/store` call, passes after revert (proven).
- **Local hub (docker), ground truth:** `create -y` (via `POST /api/v1/boxes`) → `ls` shows it → `pause`/`unpause` (verified by `docker inspect`) → `shell` (IO plane) → `destroy` (container removed + local record gone + `ls` empty).
- **Remote-shaped:** `/api/v1/boxes` returns 200 with `apiVersion: v1`; `ls -g` resolves `mode: remote` and lists cleanly with the API key; degrades with the `no-token` hint (not a crash) without it.
- Dashboard TUI isn't interactively runnable in this env (no `node-pty` prebuild); its listing + destroy logic are unit-tested and the destroy shares the verified `client.destroy` path.

Net: **-594 lines** (1224 deleted, 630 added).

https://claude.ai/code/session_01K4uGZcAXnSHDnQcCqLR6AC

---

> [!NOTE]
> **Medium Risk**
> Dashboard cloud destroy and listing merge behavior changed on critical fleet paths, though logic is unit-tested and destroy now matches the CLI's owner-first hub routing.
> 
> **Overview**
> **Step 11** finishes the `/api/v1` consolidation by deleting the CLI's second implementation on the internal relay wire and routing remaining dashboard flows through the public hub API.
> 
> The **internal admin client stack** goes away: `ControlPlaneAdminClient`, `reapOnControlBox`, registration-based `fetchHubListing`, and `hub-merge` / `listBoxesMerged`. Dead `waitForEvent` (`/admin/events`) is removed with its tests. **`HubApiClient`** is documented as the sole CLI client for box/fleet work.
> 
> The **dashboard** now calls **`listDashboardBoxes`** (`dashboard/box-list.ts`), which still merges live local `listBoxes()` probes with **`fetchBoxListing`** (`/api/v1/boxes`) so hub-only rows and rich local fields (sessions, endpoints, questions) stay available. Cloud **destroy** in the TUI uses **`withOwningHub`** + **`client.destroy`** (server-side teardown + reap), then **`removeBoxRecord`** locally—aligned with **`agentbox destroy`** instead of inline provider destroy + `/remote/boxes` reap.
> 
> A **`no-internal-wire-client`** guard test scans CLI source for `/admin`/`/remote` paths with a small allowlist (custody fallback, local queue poke, repo-installed check). Hub backend comments that referenced a mirrored CLI implementation are cleaned up; the plan doc marks Step 11 done.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b58f231da13af3ad40a497bb1c11270bea7e605. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

Addressed the High finding: dashboard cloud-destroy now routes through `withOwningHub` (owner-first + retry-the-other-hub on not_found), matching `agentbox destroy`, instead of `resolveHubApiClient(undefined)` current-config resolution. Please re-review head 3b58f231.

PR #294, opened 2026-07-30
feat(cli): docker off under a remote hub (thin CLI Step 12)

madarco:
## What

Step 12 of the [one-path plan](docs/hub-api-single-path-plan.md): when a control box is configured (`relay.controlPlaneUrl`), gate the local docker engine off — a docker box built on the laptop can't run with the laptop off, which is the whole point of a control box. Plus two structural cleanups so a docker-free host never imports docker machinery.

This is a **gate, not a removal**: with no control box configured, everything works exactly as today.

### The gate

One shared predicate — `dockerProvidersHidden(effective)` in `control-plane/remote-hub.ts` (beside `remoteHubConfigured`) — drives every site, so they can't drift:
- **`create`** (and the `claude`/`codex`/`opencode` launchers — they build boxes too, same false-coverage reason Step 13 widened its gate) and **`prepare`** refuse, naming the config key.
- **`doctor`** and the **`install`** provider picker drop docker/remote-docker rows.
- **`ls`** keeps docker boxes **listed but marked `docker (inactive)`** with a footer note — it does **not** silently hide them, so a user who configures a control box while docker boxes are running can still see and `destroy <name>` them without flipping config (destroy resolves locally from `state.json`). Silently dropping them would be the hidden-but-alive failure class rejected in Step 5's destroy bug.

New typed config key **`hub.mode: 'auto' | 'thin' | 'local'`** (default `auto`): `auto` gates when a control box is configured, `thin` forces the gate even without one, `local` keeps docker on regardless — the escape hatch every message names. A config key, not an env var (repo convention: typed key + CLI-flag precedence).

### Structural cleanups (docker-free host doesn't pull docker machinery)

- **Hub lifecycle → `@agentbox/sandbox-core`.** `ensureHub`/`getHubStatus`/`stopHub`/`resolveHubServer`/`hubRuntimeEnv`/`HUB_TOKEN_FILE` moved from `@agentbox/sandbox-docker` to `hub-lifecycle.ts`; the shared process/`/healthz` probes moved with them into `hub-process.ts` (re-exported from `relay.ts` for compat). The two docker-side niceties it used directly — Portless and the docker build context — come through a `hub-hooks.ts` seam the CLI fills at startup; the docker Portless impl is `sandbox-docker/hub-portless.ts`.
- **Severed `refreshAgentCredentialsBackup`'s docker-shared-volume reach-in** via a `credential-refresh.ts` seam (impl `dockerCredentialRefresh` in sandbox-docker, registered by the CLI), so `@agentbox/sandbox-cloud` no longer imports docker for the credential refresh.

Both mirror the existing `credential-publish.ts` one-slot-hook pattern.

### Invariants kept
- The localhost hub still binds `0.0.0.0` (carried verbatim in `hub-lifecycle.ts`) so docker boxes reach the embedded relay at `host.docker.internal:8787`.
- The custody byte-read loopback peer gate (`apps/hub/lib/peer.ts`) is untouched.

## Verified end-to-end (built CLI, ground truth)

- **Gate ON** (`relay.controlPlaneUrl` set): `create --provider docker` and `prepare --provider docker` refuse naming `hub.mode=local`; `doctor` shows 0 docker rows; `ls` marks docker boxes `docker (inactive)` with the footer note; `destroy <name>` on a marked box tore down the container (verified gone via `docker ps`) **without** flipping config.
- **`hub.mode=local`** reinstates docker everywhere (doctor shows it; create proceeds past the gate).
- **Regression half** (`relay.controlPlaneUrl` unset): `doctor` shows docker; `create --provider docker` routes through the local hub queue and brings up a real container (verified `Up` via `docker ps`) through the full pipeline (volume prep, agent-config sync, box.env) exactly as before — the hub autostarted via the **moved** `ensureHub` and resolved its Portless URL through the new seam, binding `0.0.0.0` (confirmed in the hub log). `hub restart`/`hub status` work off the sandbox-core lifecycle.
- `pnpm build`, `pnpm typecheck`, `pnpm test` (all 30 tasks) green after rebasing onto Step 11 — including Step 11's `no-internal-wire-client` guard (this change adds no `/admin/` or `/remote/` client call) and a new `remote-hub-gate.test.ts` covering the `auto`/`thin`/`local` semantics.

https://claude.ai/code/session_01C4CxXFHQjkEHduKoJnYV8N

---

> [!NOTE]
> **Medium Risk**
> Touches default create/prepare/hub startup paths and provider visibility; behavior changes only when a control box is configured or `hub.mode` is thin, with `local` as the documented escape hatch.
> 
> **Overview**
> When a **control box** is configured (`relay.controlPlaneUrl`), local **docker** / **remote-docker** are gated off by default so laptop-only sandboxes do not undermine laptop-off cloud fleets. A new **`hub.mode`** (`auto` | `thin` | `local`) controls this: `auto` gates once a control box exists, `thin` always gates, and `local` keeps docker available everywhere.
> 
> **`dockerProvidersHidden`** in `remote-hub.ts` is the single predicate behind refusals on **`create`**, **`prepare`**, and the **claude/codex/opencode** launchers (shared error text naming `hub.mode=local`), filtered **`install`** / **`doctor`** provider lists, and **`ls`** rows shown as **`docker (inactive)`** with a footer (boxes stay visible for **`destroy`**).
> 
> **Hub lifecycle** (`ensureHub`, `getHubStatus`, `stopHub`, probes) moves from **`@agentbox/sandbox-docker`** to **`@agentbox/sandbox-core`**; Portless, docker build context, and credential refresh from shared volumes are wired via **CLI-registered seams** so docker-free hosts can start the hub without importing docker. **`@agentbox/sandbox-cloud`** no longer imports docker for agent credential refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b35b625774e91167017c30b66c10bef2ca09d8c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #311, opened 2026-08-25
feat: remote-docker boxes through the control box (+ docker:hub)

madarco:
Setting up a control box used to cost you the docker-shaped box entirely: local docker is gated off (a box built on your laptop dies with your laptop), and `remote-docker` — the one provider that runs on hardware you already own — could never be driven from the hub, because its "credential" is your own `~/.ssh/config`.

This makes both work.

## What's in it

- **Share an engine with the control box.** `agentbox remote-docker share <alias>` expands the alias with `ssh -G`, mints a **dedicated key for the hub**, installs its public half on the engine over the access you already have, and sends connection + key to `POST /api/v1/hosts`. The hub probes ssh **and** docker with that key before saving, so a 201 means *it* can reach the engine — not that you could. `unshare` revokes the grant. Your own key never leaves the PC.
- **The hub can build those boxes.** A `docker:<alias>` box bind-mounts nothing and is seeded over SSH exactly like a cloud box, so the clone path now refuses only *bare* docker; the worker resolves the spec and threads the alias through as `providerOptions.remoteHost`. Creates from the PC route to the control box when it knows the alias, and fall back locally with a reason when it doesn't.
- **`docker:hub` — the control box's own engine.** The deploy mints a key on the VPS, trusts it for root, and registers `hub → host.docker.internal` (compose gains the matching `extra_hosts`). Setup registers it on the PC too and, if your default was still plain `docker`, moves `box.provider` to `docker:hub` — so plain `agentbox create` keeps making the same kind of box, on a machine that stays on. `hub destroy` puts it back.
- **Docker-29 fix (prerequisite).** The remote build streamed the context to `docker build -` *with* `-f`, which buildx rejects outright (`ambiguous Dockerfile source`) — every bake and every registry-miss create was dead on a current engine. It now stages the context in a temp dir and builds from a directory context, with an absolute `-f` (buildx resolves `--file` against the *client's* cwd).
- The docker gate also stops firing for a control box that IS this machine (`hub expose` / a loopback `set-url`), which was switching local docker off for no reason.
- `parseProviderSpec` moves to `@agentbox/config` so the CLI and hub read `docker:<host>` the same way instead of via a second regex.

## Verified live

Against the Hetzner control box (docker 29.7.2), end to end: forced `--build` bake on its engine; `hub` registered on both sides; `agentbox create` from the PC with the `docker:hub` default → **built by the control box**, container on its engine, registered as `remote-docker` with `sandboxId: hub/agentbox-<name>`; the same create via `POST /api/v1/boxes {provider:"docker:hub"}` (what the web UI sends), with `Docker (hub)` in the picker; and an in-box `agentbox-ctl git push` served by the hub relay's scratch-repo path. Boxes destroyed, config restored.

Two sharp edges it surfaced are pre-existing and provider-agnostic — hub-created boxes clone shallow (which the bundle push can't complete from) and a hub box's `webUrl` is the hub container's loopback port. Both are written up in `docs/remote-hub-backlog.md`.

Plan + phases: `docs/remote-docker-hub-plan.md`.

---

> [!NOTE]
> **Medium Risk**
> Touches control-box create routing, hub host registry, and SSH key minting/sharing; mistakes could route creates to the wrong engine or leave stale share state, though 4xx rollbacks and probes limit blast radius.
> 
> **Overview**
> **Remote-docker on the control box.** PCs can **share** a registered engine with the hub (`agentbox remote-docker share|unshare`, auto on `remote-docker add` unless `--no-share`): the CLI expands `ssh -G`, mints a hub-only key, and `POST /api/v1/hosts` with `connection` + `identity`; the hub probes SSH/docker before saving. Create/prepare routing now treats **shared** `docker:<alias>` like a hub-routable engine (wire `provider: docker:<alias>`, pass `remoteHost` through agent launchers and `create`), while **bare** local docker stays gated when a control box is configured.
> 
> **`docker:hub`.** After setup/deploy/update, `finalizeControlBoxState` re-syncs shared hosts and runs `ensureHubDockerTarget` (local alias `hub` → deploy SSH, optional `box.provider` flip to `docker:hub` when the hub knows that engine). `set-url` / `unset-url` and hub destroy clean that up. Docker refusal logic adds exceptions: co-located hub (`hub expose`), and remote-docker engines the hub can run.
> 
> **Hub worker/API.** Control-plane create accepts `docker:<alias>` (not only cloud); the worker parses specs via shared `parseProviderSpec` from `@agentbox/config` and passes `remoteHost` into the provider. Host registry stores managed keys and dials shared engines with them; compose adds `host.docker.internal` for the hub’s own Docker.
> 
> **Other.** `runPrepare` throws `UserFacingError` instead of `process.exit` for embedders; remote Docker **build context** is staged on the engine (Docker 29 / buildx fix per changelog). Docs/CLAUDE note hub `/api/v1` as the single execution path for real work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 931b37ac0c0a46bde47d81d44ca5bb0e7b83755a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #313, opened 2026-08-26
fix(agent): read a box's in-TUI prompts from the hub that owns it

madarco:
Fixes the Bugbot finding on #312 ("Hub boxes miss in-TUI approvals"), which is real.

## The bug

`readBoxStatus` reads `~/.agentbox/boxes/<box>/status.json` — a file written by **whichever relay the box reports to**. A box a control box created reports there, so the file is on that machine's disk and never exists on your laptop. The codebase already states this invariant twice (`prompt-client.ts:18`, `hub-list.ts:6`).

The `/api/v1` conversion moved `agent state`, `wait-for` and `get-plan-question` onto `client.getAgentState(box.id)`. Two call sites were left on the local read:

- `gatherApprovals` — the plan / question / tool-permission rows of `agent approvals`
- `approveInTui` — the race guard in `agent approve`

So for a control-box box, `agent approvals` showed only host-action rows (those *do* come from the hub) and a box parked on a plan approval reported **"agent not parked on a prompt"**. `agent approve` then re-read the same absent file and refused the prompt as already changed.

Same root cause as a bug already fixed in this range — the attach footer, which "read a status file only the local relay writes".

## The fix

Both call sites go through the hub that owns the box. `approveInTui` uses the existing `fetchAgentClaude` helper (so it inherits the `HUB_ERROR` / `HUB_NOT_FOUND` handling); `gatherApprovals` uses `source.client`, the same client its host-action half already used — so there is one decision about which hub owns the box, not two that could disagree.

**No change for a local box.** The hub's `getAgentState` reads that very same `status.json` on its own disk, falling back to its in-memory status store — a superset of what the direct file read returned.

One behavioural consequence worth calling out: the two halves of `approvals` are no longer independent, since both now hit the same hub. The old code assumed the in-TUI half was "entirely independent" and reported only a mailbox failure. Each half is now tracked and named separately, because a half we could not read must never render as "nothing pending".

## Testing

New `apps/cli/test/agent-approvals-source.test.ts` (6 tests, pure — a stub client, no fs or network): a hub box's plan prompt surfaces; a null snapshot yields no row *with both halves answered*; host-action rows are merged and filtered to the box; and each half degrades on its own, naming the other. Test 1 fails without the fix.

Full CLI suite (2520), `pnpm typecheck` and `pnpm lint` green.

Live, against the real control box at `46.225.235.16`:

- `GET /api/v1/boxes/<id>/agent` over the WAN with the Bearer key returns `{"claude":{"state":"unknown","updatedAt":null,"sessionRunning":false}}` for a hub-created box — the exact `BoxStatusClaude` shape the new code reads, and a structured `not_found` for an unknown id.
- That same box has **no `~/.agentbox/boxes/<seg>/` directory at all** on this laptop, so the old read returned null unconditionally. The bug was total for such boxes, not intermittent.

The snapshot→row transformation itself (`inTuiKind`/`mintTuiId`) is pure and covered by the new tests plus the existing `agent-answer.test.ts`.

---

> [!NOTE]
> **Medium Risk**
> Changes hub routing and agent snapshot/approval reads across local, control-box, and remote-docker setups; mistakes would mis-list or mis-approve prompts, but behavior is covered by new tests and stays CLI-side.
> 
> **Overview**
> Fixes **control-box boxes** where `agent approvals` / `approve` missed plan, question, and permission rows because the in-TUI half still read local `status.json` (only the local relay writes it). **Both halves now use the owning hub’s** `getAgentState`, same as host-action approvals.
> 
> All **`agent` snapshot readers** (`state`, `wait-for`, `get-plan-question`, and the in-TUI race guard) share **`resolveBoxPromptSource`** instead of `withOwningHub`, so a locally-adopted hub box no longer gets stuck on this laptop’s `{claude: null}` while the control box holds the real snapshot. **`boxAnswersOnLocalHub`** replaces inline `provider === 'docker'` checks so **`remote-docker`** boxes register locally vs on the control box (`docker:hub`) route to the right hub.
> 
> **Partial failures** are tracked separately (`missingHalves`, `tuiError`): a half that couldn’t be read no longer prints “nothing pending”, and missing **`AGENTBOX_HUB_API_KEY`** is reported consistently instead of looking like an empty snapshot.
> 
> Adds **`agent-approvals-source.test.ts`** (stub hub client) and **CHANGELOG** entries for these fixes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a99010f618b26dc8e1ed913ad1f1ddf0c83fcb29. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -189,6 +189,12 @@ CLI, not the raw commits.
 
 ### Fixed
 
+- **`agentbox agent approvals` missed the in-TUI prompts of a control-box box.**
+  Plan, question and tool-permission rows were read from a status file only the
+  local relay writes, so a box parked on a plan approval reported "agent not
+  parked on a prompt" — and `agent approve` refused the same prompt as already
+  changed. Both now read the snapshot from the hub that owns the box, and a half
+  that could not be read is reported instead of rendering as "nothing pending".
 - **Remote Docker builds work on Docker 29.** The box image was streamed to
   `docker build -` alongside `-f`, which Docker 29's buildx rejects outright
   ("ambiguous Dockerfile source") — that broke every bake and every registry-miss
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -195,6 +195,11 @@ CLI, not the raw commits.
   parked on a prompt" — and `agent approve` refused the same prompt as already
   changed. Both now read the snapshot from the hub that owns the box, and a half
   that could not be read is reported instead of rendering as "nothing pending".
+- **`agent state`, `wait-for` and `get-plan-question` reported no snapshot for a
+  box the control box created.** They asked this machine's hub first, and a hub
+  box the local registry knows answers "no snapshot" rather than "no such box" —
+  so the fallback to the control box never happened. All five `agent` readers now
+  resolve the owning hub the same way.
 - **A remote-docker box's approvals were looked for on the wrong hub** when a
   control box was configured. It registers with the local relay, but an inline
   provider check treated only `docker` as local, so its host-action mailbox was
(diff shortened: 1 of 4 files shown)


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #315, opened 2026-08-26
feat(config): say when a box size will be ignored, before it is

madarco:
## The bug

`box.sizeDaytona = 4-8-30` appeared to have no effect on boxes created with `-i`.

The value is plumbed correctly all the way to `req.size`. The daytona backend then **discards it by necessity** — with `box.imageDaytona` set, creates take the snapshot path, and Daytona rejects per-create resources:

```ts
// packages/sandbox-daytona/src/backend.ts:331
// Daytona rejects `resources` on the snapshot path — the snapshot's own params encode them.
delete snapshotParams.resources;
```

The backend already warned, and it fired on every run — into `~/.agentbox/logs/queue-<id>.log`, because a detached `-i` job's `onLog` goes to the job log. Nobody opens that for a box that came up fine, so the setting looked silently ignored.

**E2B has the identical defect** (`packages/sandbox-e2b/src/backend.ts:161-180`). hetzner and vercel apply `size` live and are unaffected.

## The change

**`ProviderModule.sizeIgnoredReason?(size): string | null`** — optional, beside `currentBaseFingerprintLive`. Per-provider rather than shared for the reason already written at `doctor.ts:86`: each parses its own size grammar (`cpu-mem-disk` vs `cpu-mem`) and reads the baked value from its own prepared-state shape. Implemented by lifting the comparison already inline in each backend, so message and behaviour cannot drift.

Contractually **pure and local** — one JSON read, no network, no credentials — because it runs on every size `config set`.

Called where someone is looking:
- `agentbox config set box.size` / `box.size<Provider>` (`warnOnSet`, which already exists for `queue.openIn`). The generic key defers to the *effective* provider so a global default doesn't warn about five unused backends.
- `submitQueueJob`, before enqueueing — resolving the same effective size the worker will, so `-i` warns in the submitter's terminal.

Backend warnings stay: they are ground truth at provision and still cover hub-driven creates that never touch this CLI.

**Daytona records what a snapshot actually has** via a new `extras.effectiveSize`, so the warning names real numbers instead of "the default size". Deliberately *not* `extras.size` — that field feeds both `defaultSnapshotName` and `preparedMatches`, so writing a default into it would rename every existing snapshot and re-bake the world. `preparedMatches` is untouched; there is no rebake risk.

Nothing is gated. A mismatched size is legal, and every failure path degrades to silence.

## Verified

Unit: 13 new tests (daytona 6, e2b 4, CLI key-mapping 3). Full suite + `pnpm typecheck` green.

Live, on this machine:

| case | result |
|---|---|
| `config set box.sizeDaytona 4-8-30` — the reported case | warns, names the re-bake command |
| `config set box.sizeE2b 4-8` | warns |
| `config set box.sizeHetzner cx23` — control, applies live | silent |
| `config set box.sizeVercel 4` — control, applies live | silent |
| `claude -i --provider daytona` | **warns in the submitting terminal** — the symptom |

The `-i` submit ran with the queue gated to 0 concurrency and the job cancelled, so no cloud box was created.

**Not verified live:** the `effectiveSize` recording, which only runs during a real `agentbox prepare --provider daytona --force`. That re-pins `box.imageDaytona` globally and there is a live daytona box in use, so I did not re-bake unprompted. Covered by unit tests; worth one real bake before relying on the numbers in the message.

## SDK

`ProviderModule` is re-exported by `@madarco/agentbox-provider-sdk`. The member is optional, so this is additive — bumped to **2.6.0**, no `SDK_API_VERSION` change. Needs a separate republish; the CLI publish does not cover it.

---

> [!NOTE]
> **Low Risk**
> Advisory-only CLI notes and an optional provider hook; no create/config blocking. Daytona adds optional prepared-state metadata during bake, which is additive and covered by tests.
> 
> **Overview**
> **Daytona and E2B fix VM resources at bake time**, so a `box.size` / `box.size<Provider>` or `--size` that disagrees with the prepared base is dropped at create. Warnings used to land only in detached queue logs; this PR surfaces them where users actually look.
> 
> The CLI adds **`size-advisory`** helpers that load the provider and call a new optional **`ProviderModule.sizeIgnoredReason`** hook (`@madarco/agentbox-provider-sdk` **2.6.0**). **`agentbox config set`** on size keys prints a `note:` after a successful write (per-provider keys target that provider; generic `box.size` uses the effective `box.provider`). **`submitQueueJob`** prints the same note **before** enqueueing `-i`, resolving size the same way as the worker (flag → per-provider key → `box.size`).
> 
> **Daytona** implements the hook against prepared-state bake metadata and records **`extras.effectiveSize`** during `prepare` so messages cite real CPU/memory/disk instead of “the default size,” without changing how `extras.size` drives snapshot naming. **E2B** implements the same hook against its template bake size. Docs and changelog describe the behavior; unit tests cover key mapping and both providers’ reason strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca5187d93c51037757894aa53ab5389b3dea9695. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #316, opened 2026-08-26
fix(e2b): pause at the session cap instead of being killed by it

madarco:
## The finding

E2B boxes were dying at ~60 minutes. The keepalive was **not** missing — `cloud-keepalive.ts` already sweeps every 60s and calls `Sandbox.setTimeout` for e2b, which is why boxes reached 60 min rather than lapsing at the 45-min create TTL.

60 minutes is E2B's **plan cap**, stated verbatim in the vendored SDK typings (`e2b@2.27.1`, `SandboxConnectOpts.timeoutMs`):

> Maximum time a sandbox can be kept alive is 24 hours for Pro users and **1 hour for Hobby users**.

No ping gets past that. So the fix is to survive the cap, not to outrun it.

## What changed

**1. The cap pauses instead of killing.** `Sandbox.create` now passes `lifecycle: { onTimeout: 'pause', autoResume: true }` (SDK default is `'kill'`). Hitting the cap freezes the box — filesystem *and* memory — and the next inbound request revives it. The hour mark becomes a stall, not a lost box.

**2. E2B becomes an inactivity-model backend.** Auto-resume means any request revives a paused box, and the relay's `CloudBoxPoller` long-polls every cloud box forever — the exact trap `CloudBackend.timeoutModel` documents for Daytona. So e2b declares `timeoutModel: 'inactivity'`, and the keepalive's idle-pause path now **also stops that box's poller** (`RelayServerHandle.stopCloudPoller`), without which the pause is undone on the next long-poll.

**3. Resume re-arms the session window.** `Sandbox.connect` *always* posts a sandbox deadline and defaults it to **5 minutes** (`connectSandbox`, `dist/index.js:4124`), so every resumed box was living on borrowed time and nothing re-applied `box.e2bTimeoutMs`. `reEnsureCloudBox` now calls `renewTimeout` on wake (fixes Vercel too), and all five `Sandbox.connect` sites pass an explicit `timeoutMs`.

**4. Deadline bookkeeping survives a resume.** New `cloud.sessionDeadlineEpochMs` records the deadline the provider actually applied; the keepalive seeds from it instead of deriving `createdAt + sessionTimeoutMs`, which is stale after any pause/resume and points into the past after a relay restart mid-session.

Plus: `e2bTimeoutMs` gets the `"minimum": 1` its two sibling keys already have (it was the only timeout key where `0` flowed straight into `Sandbox.create`), and the config docstring — which claimed e2b "auto-pauses on inactivity" when it actually killed on an absolute timer — is now true.

## Verification

Unit: 5 new cases in `cloud-keepalive.test.ts` (28 total) — deadline seeding prefers the recorded value, falls back for pre-feature records; `stopPoller` fires once after a successful pause, not when pause throws, and a failing `stopPoller` doesn't re-arm the box. Full suite 1316 passed, `pnpm typecheck` + `pnpm lint` green.

Live, against a real E2B box:

| check | result |
|---|---|
| policy applied at create | `lifecycle: {"onTimeout":"pause","autoResume":true}` via `Sandbox.getInfo` |
| box usable | `agentbox shell` → `vscode`, `/workspace` populated, on branch `agentbox/e2bttl` |
| **resume TTL** | after `stop` + `start`: **59.6 min** left, not 5 — this is the regression the fix targets |
| recorded deadline | `cloud.sessionDeadlineEpochMs` written on resume |
| policy survives resume | still `onTimeout: pause` after wake |
| cleanup | destroyed; `Sandbox.list` shows no orphans |

**Not live-verified:** the idle-pause-plus-poller-stop path end to end — that needs a box sitting idle for a full `box.e2bTimeoutMs`. It's covered by unit tests and the pause primitive itself is unchanged.

## Note for review

`BoxRecord` / `CloudBoxFields` are re-exported by `@madarco/agentbox-provider-sdk`, and this adds an optional field to the cloud block. Additive, so no `SDK_API_VERSION` bump — but the emitted `.d.ts` moves, so the SDK wants a republish to stay in sync.

https://claude.ai/code/session_01SKLmAbYKtGVKNzKG3j9eqp

---

> [!NOTE]
> **Medium Risk**
> Changes cloud session lifecycle, relay admin API, and keepalive/poller coupling for E2B (and resume timeout re-arm affects other clouds via `reEnsureCloudBox`); mis-stopped pollers could leave boxes unreachable until re-register.
> 
> **Overview**
> E2B sandboxes were dying at the ~1 h Hobby plan cap because the SDK defaults to **`onTimeout: 'kill'`**. Creates now use **`lifecycle: { onTimeout: 'pause', autoResume: true }`**, so hitting the cap freezes state instead of destroying the box; docs and **`box.e2bTimeoutMs`** copy reflect pause/resume and re-arm on wake.
> 
> Because auto-resume revives on any traffic, E2B is treated like Daytona as an **inactivity-timeout** backend: the keepalive loop idle-pauses after **`box.e2bTimeoutMs`** and must **stop the relay `CloudBoxPoller`** (new **`POST /admin/stop-poller`**, **`stopBoxPollerOnRelay`**, wired from **`pause`/`stop`** and keepalive). Without that, long-polls immediately undo a pause.
> 
> **Resume** re-applies the configured session window: every **`Sandbox.connect`** passes explicit **`timeoutMs`** (SDK otherwise defaults to 5 min), **`reEnsureCloudBox`** calls **`renewTimeout`** on wake, and **`cloud.sessionDeadlineEpochMs`** records the last applied deadline so keepalive seeds correctly after pause/resume or relay restart. **`e2bTimeoutMs`** gets schema **`minimum: 1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35afc1fa79a0ffab967eb851c4facd0e6fd15ba6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/content/docs/configuration.mdx
+++ b/apps/web/content/docs/configuration.mdx
@@ -219,7 +219,7 @@ See [teleport a project](/docs/teleport-a-project), [environment](/docs/environm
 | ------------------------- | ------ | ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `box.vercelTimeoutMs`     | int    | `2700000` (45 min)  | max session length before auto-snapshot; persistent mode auto-resumes                                                                                                                         |
 | `box.vercelNetworkPolicy` | string | empty (`allow-all`) | egress lock: `allow-all`, `deny-all`, or a comma-separated domain allowlist (`github.com,*.npmjs.org`)                                                                                        |
-| `box.e2bTimeoutMs`        | int    | `2700000` (45 min)  | session timeout a new `--provider e2b` box is created with before E2B auto-pauses it on inactivity; the host keepalive holds it open while the agent works (Hobby caps total session at ~1 h) |
+| `box.e2bTimeoutMs`        | int    | `2700000` (45 min)  | session window a `--provider e2b` box is created with and re-armed with on resume; the host keepalive holds it open while the agent works, and pauses the box once the agent has been idle that long. Boxes pause rather than die at the window — and at E2B's own cap (~1 h Hobby, 24 h Pro) — keeping filesystem and memory |
 
 See [Vercel](/docs/vercel).
 
--- a/apps/web/content/docs/e2b.mdx
+++ b/apps/web/content/docs/e2b.mdx
@@ -3,7 +3,7 @@ title: E2B
 description: Run your agents in a fast E2B cloud sandbox with public preview URLs and free pause/resume
 ---
 
-Run your agents in a fast E2B cloud sandbox with a public HTTPS preview URL and free pause/resume — no local Docker needed. Each box is a remote E2B sandbox you drive with the same `agentbox` commands as a local box. It supports [Docker-in-Docker](/docs/docker-in-docker) (the engine is baked into the base) but the Hobby tier caps sessions at one hour — and it's the only provider that builds its base straight from a Dockerfile.
+Run your agents in a fast E2B cloud sandbox with a public HTTPS preview URL and free pause/resume — no local Docker needed. Each box is a remote E2B sandbox you drive with the same `agentbox` commands as a local box. It supports [Docker-in-Docker](/docs/docker-in-docker) (the engine is baked into the base), and it's the only provider that builds its base straight from a Dockerfile. The Hobby tier caps a session at one hour, but reaching that cap pauses the box rather than destroying it — the next command wakes it up where you left off.
 
 Switch per box with `--provider e2b`, or pin it project-wide with `box.provider: e2b` in [`agentbox.yaml`](/docs/agentbox-yaml). Comparing options? See [local-docker](/docs/local-docker), [hetzner](/docs/hetzner), [daytona](/docs/daytona), and [vercel](/docs/vercel).
 
@@ -87,7 +87,7 @@ agentbox e2b claude
 | Region            | SDK-chosen (transparent)                                                                                                                    |
 | Disk / RAM / vCPU | Template-level — baked at `Template.build()` time via `prepare --provider e2b --size <cpu-memory>`; per-create `--size` warns if it differs |
 | Exposed ports     | Any port via `sandbox.getHost(port)`; WebProxy runs on 8080                                                                                 |
-| Session length    | 1 hour (Hobby) — the attach helper caps at **55 minutes** for headroom; auto-renewed (up to the plan cap) while the agent is working        |
+| Session length    | 1 hour (Hobby) / 24 hours (Pro) — auto-renewed while the agent is working; at the cap the box **pauses and auto-resumes**, it is not killed. The attach helper caps at **55 minutes** for headroom |
 | Nested containers | Yes — the docker engine is baked into the prepare template; `dockerd` auto-launches on create/resume                                        |
 | SSH               | none — attach is an SDK-streaming PTY bridge over `pty.create`                                                                              |
 
--- a/docs/cloud-providers.md
+++ b/docs/cloud-providers.md
@@ -891,14 +891,21 @@ brief:
   next inbound request revives it with a fresh window. The hour mark is a brief
   stall, not a lost box. (The SDK default is `onTimeout: 'kill'` — leaving it
   there was why boxes were observed dying at exactly 60 min.)
-- **Host-owned idle policy (`timeoutModel: 'inactivity'`).** Auto-resume means
-  any inbound request revives a paused box, and the relay's `CloudBoxPoller`
-  long-polls every cloud box forever — so E2B falls into the same trap Daytona
-  does (see §3, "Idle handling") and the host must do the stopping itself. The
-  keepalive loop pauses a box whose agent has been idle for a full
-  `box.e2bTimeoutMs`, **and stops that box's poller** (`stopCloudPoller`) so the
-  pause actually sticks. A later wake re-registers the box and a fresh poller
-  starts.
+- **Any pause must also stop the box's poller.** Auto-resume means inbound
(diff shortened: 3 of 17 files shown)


madarco:
Correction to the PR description: I claimed the resume re-arm "fixes Vercel too". It did not — it introduced a bug there, now fixed in e4c8d87f.

The re-arm passed `current: now`, i.e. "the session has zero time left". Vercel's `extendTimeout` is **additive**, so that stacked a full 45-minute window on top of whatever the box already had — and since `reEnsureCloudBox` also runs on `reconnect` against a live box, it stacked again on every call. It now passes the recorded deadline when that's still in the future, and skips the renew entirely when the box already has at least the window we'd be asking for.

No behavior change on e2b: its `setTimeout` is absolute-from-now and ignores `current` by design, which is why the live verification above still holds.

madarco:
## Found a regression this PR would have shipped — fixed in 35afc1fa

`autoResume: true` means inbound traffic revives a paused box. The relay long-polls every cloud box's preview URL every ~25s. So **`agentbox stop` on an e2b box did not stick** — the poller resurrected it within seconds.

Measured, with the poller running:

```
stopped e2bpause
22:37:59 state: running     22:39:01 state: running
22:38:20 state: running     22:39:21 state: running
22:38:40 state: running     22:39:42 state: running
```

Controlled comparison — same box, poller stopped first, paused via the SDK:

```
22:40:18 state: paused      22:41:19 state: paused
22:40:39 state: paused      22:41:40 state: paused
22:40:59 state: paused
```

I had wired the poller stop into the keepalive's *idle* pause but not the *user-initiated* one. Added `POST /admin/stop-poller` — which unlike `/admin/forget-box` keeps the registration and status, so `list` and the hub keep showing the box — and called it from `provider.pause` and `provider.stop`. The wake path already re-registers the box and starts a fresh poller, so there's no restart counterpart needed.

Re-verified end to end after the fix:

```
stopped e2bpause
22:49:07 state: paused      22:50:08 state: paused
22:49:27 state: paused      22:50:28 state: paused
22:49:48 state: paused      22:50:49 state: paused
```

Then `agentbox start` → `RESUME_OK` / `vscode`, destroyed, zero orphans.

### Process note

My first attempt at verifying this fix reported failure, and I nearly concluded the fix didn't work. The cause was the stale-bundle gotcha in `CLAUDE.md`: the **hub** owns `:8787` and had been spawned from the staged bundle, so `agentbox relay restart` never touched it — `hub status` said `stale: rebuilt since this hub started`. The rebuild + `AGENTBOX_HUB_BIN` override was needed. Worth knowing that a relay-route change tested through a running hub silently exercises old code.

This does not affect the resume-TTL verification reported earlier: that path runs in the CLI process, which was freshly built.

Tests: 2 new cases for the route (registration survives; missing `boxId` → 400). Full suite green, typecheck + lint clean.

madarco:
bugbot run

madarco:
## Measured both findings before acting. One is confirmed, one is refuted.

### Finding 1 (High) — refuted

The premise was that the absolute timer pauses an idle box minutes after the agent idles and the still-running poller revives it in a loop. Neither half reproduces on a live box.

**The keepalive never shortens a deadline, only pushes it out**, so an idle box keeps riding its original create deadline — there is no early pause. And with the poller running the deadline never expires at all; it gets pushed back to a full window:

```
07:29:37 +1.7m  running  ends in 0.2m     <- about to expire
07:29:57 +2.0m  running  ends in 59.9m    <- fresh full window
07:36:22 +8.4m  running  ends in 53.5m    <- decaying normally since
```

Zero state flips, at 20s sampling and again at 2s sampling across a deliberately compressed deadline (60 samples, `flips: 0`). No oscillation exists.

What actually stops the box is the host idle-pause, and it works:

```
cloud-keepalive: paused idle box bff1a1dd9 (e2b) after ~23m idle
  — e2b's own idle timer never fires while we poll it
```
followed by 14/14 `paused` samples.

That last line is also the direct evidence that **`timeoutModel: 'inactivity'` is the correct label**: our own polling holds the deadline open exactly as it does on Daytona. E2B's raw timer is absolute, but no box under polling ever reaches it. No change made for this finding.

### Finding 2 (Medium) — confirmed, and the live numbers understated it

Timing `POST /admin/stop-poller` against a real box looked harmless (~0.03s, 6/6 rounds) — but a healthy bridge answers instantly, so the loop is almost never mid-request. Against a bridge that accepts the poll and never answers, i.e. a genuine long-poll:

```
>>> stop() with a poll in flight took 25997ms
```

So the revival window is a full 26 seconds, and it opens precisely when a box is slow or wedged. Fixed in a5050470:

- `CloudBoxPoller` tracks the in-flight request and `stop()` destroys it — **26s → 304ms** in the new test.
- The resulting socket error is treated as a clean shutdown, so it neither logs `poll error` nor falls through to backoff / preview-URL recovery (both asserted).
- Both pause paths now stop the poller **before** pausing, so the abort happens while the box is still up.
- If the pause then fails, `provider.pause` re-registers the box rather than leaving it running unpolled.

### Verification

Full suite green (relay 451, CLI 1316, sandbox-cloud 180, ctl 353), lint + typecheck clean. End to end on a fresh box with the rebuilt hub: `agentbox stop` returned in **1.4s** and held `paused` 6/6 samples; `start` → `FINAL_OK` / `vscode`; destroyed; zero orphans.

Docs now describe the measured mechanism rather than the one I had assumed.

### Not a bug, but worth a decision

An idle E2B box bills until `box.e2bTimeoutMs` (60 min) after last activity, and a full wake measured **6.2s**. Shortening that window would cut idle cost substantially for a ~6s wake. Left alone here since it is a policy change, not one of these findings.

PR #317, opened 2026-08-27
fix(gh): point the host gh at the repo's GitHub host (GHES support)

### LINKED ISSUE #242: PR operations fail on GitHub Enterprise Server repos
`agentbox-ctl git pr create` fails when the workspace repo is on a GitHub Enterprise Server instance. The relay spawns `gh` without `GH_HOST` set, so `gh` always targets `github.com` regardless of where the remote actually is.

## Steps to reproduce

In a box whose remote is a GHES instance (e.g. `https://github.example.com/org/repo`):

```bash
agentbox-ctl git pr create --title "my PR" --body ""
# → error: gh not authenticated on host
```

Read-only ops (`pr view`, `pr list`) fail the same way.

## Root cause

`runHostGh` in `packages/relay/src/gh.ts` spawns `gh` with `env: process.env`:

```typescript
const child = spawn('gh', args, {
  cwd,
  env: process.env,   // relay's host env — GH_HOST never set here
  ...
});
```

The relay inherits the host's login environment where `GH_HOST` isn't set, so `gh` falls back to `github.com`. The `assertGhReady` auth check is already host-agnostic (no `--hostname github.com` flag, any authed host passes), so the gap is just that `runHostGh` never tells `gh` which host to use.

## Suggested fix

Derive the hostname from the worktree's git remote and inject it as `GH_HOST`:

```typescript
function ghHostFromCwd(cwd: string): string | null {
  try {
    const remote = execSync('git remote get-url origin', { cwd }).toString().trim();
    const match = remote.match(/^https?:\/\/([^/]+)/) ?? remote.match(/^git@([^:]+):/);
    const host = match?.[1];
    return host && host !== 'github.com' ? host : null;
  } catch {
    return null;
  }
}

export function runHostGh(args, cwd, timeoutMs = GH_RPC_TIMEOUT_MS) {
  const ghHost = ghHostFromCwd(cwd);
  const env = ghHost ? { ...process.env, GH_HOST: ghHost } : process.env;
  const child = spawn('gh', args, { cwd, env, ... });
  ...
}
```

`github.com` repos are unaffected. `GH_HOST` is only injected when the remote host is something else.

## Environment

- agentbox: 0.24.6
- Provider: docker

madarco:
Fixes #242 — `gh pr`/`gh run`/`gh api` from a box fail when the repo lives on a GitHub Enterprise Server instance.

## What was wrong

The relay proxies a box's `gh` ops to the **host's** `gh`, but never told it which GitHub host to use:

- `runHostGh` spawned `gh` with a bare `process.env` — no `GH_HOST`, so `gh api repos/o/r/pulls/1/comments` went to api.github.com whatever the remote said.
- The readiness probe ran a bare `gh auth status`, which exits non-zero when **any** configured host has a stale token — so one dead github.com entry masked a perfectly good enterprise login and surfaced as our own `gh not authenticated on host` (exit 4), which is exactly what the issue reports.
- `GH_TOKEN`/`GITHUB_TOKEN` short-circuited the probe, but gh reads `GH_ENTERPRISE_TOKEN`/`GITHUB_ENTERPRISE_TOKEN` for non-github.com hosts.

## What it does now

`resolveGhTarget(originUrl)` decides the host once per call, from the box's **registered** origin — never from anything the box says, since it selects which instance the host's own authenticated `gh` gets pointed at (the issue's suggested patch shells `git remote get-url` inside the relay; that would let a box rewrite its own origin and aim the host's credentials elsewhere).

| Origin | Result |
|---|---|
| any github.com form, unparseable, local path | no `GH_HOST`, bare `auth status` — **byte-identical to today, zero extra subprocesses** |
| `https://ghe.corp/o/r` | `GH_HOST=ghe.corp`, probe scoped with `--hostname` |
| `git@ghe.corp:o/r` | `ssh -G` expands the host first, then as above |
| `git@github.com-work:o/r` (ssh alias → github.com) | alias expanded → default path, **no bogus `GH_HOST`** |
| enterprise host, not authed | exit 4 naming the host: `gh auth login --hostname ghe.corp` |
| enterprise host, `ssh -G` unavailable | falls back to the old host-agnostic behavior rather than break a working setup |

A wrong `GH_HOST` hard-fails even inside a valid checkout (*\"none of the git remotes configured for this repository correspond to the GH_HOST environment variable\"*), so the `ssh -G` expansion — the same resolution git and gh do — is what keeps multi-account alias remotes working. It reuses `resolveSshConfigTarget` from `@agentbox/sandbox-core`. Auth verdicts, the `gh --version` probe and ssh expansions each cache 60s.

## Also (wider scope, requested)

Host-side `gh auth token` callers were equally github.com-only:

- **GHCR login** is now pinned to `--hostname github.com`. A bare `gh auth token` on a GHES-configured machine returns an *enterprise* token, which that code then fed to `docker login ghcr.io`.
- **Control-box token** (`hub.gitAuth=gh`) and the **credentials gate** (`--dangerously-with-credentials`) ask about the project's own origin host, including the right REST base for identity/scope lookups (`<host>/api/v3`).

## Verification

- `pnpm test` / `pnpm typecheck` / `pnpm lint` green; 13 new unit tests (8 in `server.test.ts` asserting the exact `GH_HOST` and argv the child `gh` receives, via the PATH stub, plus an `ssh -G` stub).
- Live no-regression on this host: relay rebuilt + restarted, `gh pr list` through the relay from two docker boxes (an `https://` and a `git@github.com:` origin) still returns real PRs.
- Enterprise rows have no live GHES available — they are covered by the stubbed tests and by running `resolveGhTarget` against the real `gh`/`ssh` on this machine (github.com origins untouched; a GHES origin fails with the host-named message).

Docs updated in the same change: `apps/web/content/docs/sync-and-git.mdx` (Pull requests) and `docs/architecture.md`.

https://claude.ai/code/session_018vPQLs1xvqUJiACLPMBHY9

---

> [!NOTE]
> **Medium Risk**
> Touches host credential selection and relay `gh` proxying (security-sensitive), but github.com paths are explicitly preserved and host choice is pinned to registered origin metadata.
> 
> **Overview**
> Fixes box-proxied **`gh pr` / `gh run` / `gh api`** on **GitHub Enterprise Server** by deriving the target host from the box’s **registered** `originUrl` (not box-supplied remotes), then setting **`GH_HOST`** on host `gh` spawns and scoping readiness to **`gh auth status --hostname <host>`**. **github.com** repos keep the previous behavior (no `GH_HOST`, generic auth probe). SSH-shaped remotes expand through **`ssh -G`** so `~/.ssh/config` aliases don’t become bogus `GH_HOST` values; enterprise token env checks distinguish **`GH_ENTERPRISE_TOKEN`** from **`GH_TOKEN`**.
> 
> **Also:** hub token discovery, **`--dangerously-with-credentials`**, and GHCR **`gh auth token`** now use **`--hostname`** matching the project’s **`origin`** (plus **`githubApiBase`** for GHES REST). Docs cover enterprise **`gh auth login --hostname`** on the host.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d99659d3f6c585004763ca0be4f3a67fd858b74a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Bugbot's finding was real and is fixed in de622ee0.

`originGitHost` was passing the **raw** remote hostname to `gh auth token --hostname` and to git credential fill, so a `git@github.com-work:owner/repo` remote asked for a credential for a host that doesn't exist — a regression against the previous bare `gh auth token` / hardcoded `host=github.com`. The new `apps/cli/src/lib/git-host.ts` expands the host through `ssh -G` (same resolution the relay's `resolveGhTarget` does), and both callers use it.

One deliberate asymmetry in the with-credentials gate: only the **gh lookup** follows the alias. The `~/.git-credentials` line stays keyed by the host git will actually put in the remote URL, otherwise the helper would never match it.

Covered by `apps/cli/test/git-host.test.ts` (4 cases, ssh config stubbed).

Issue #302, opened 2026-08-08
agentbox relay start reports success when the relay failed to bind


aspiers:
## Summary

`agentbox relay start` prints `relay running on http://127.0.0.1:8787` even when the relay failed to bind and exited immediately. It reports the *intended* URL rather than a verified listener, so a hard startup failure looks like success.

## Version

- `@madarco/agentbox` 0.27.2 (relay commit `c9424ad0`)
- Linux, docker + e2b providers configured

## Reproduction

With another process already listening on 8787:

```console
$ agentbox relay start
◇  relay running on http://127.0.0.1:8787

$ agentbox relay status
relay: not running
  log:    /home/adam/.agentbox/relay.log

$ tail -1 ~/.agentbox/relay.log
agentbox-relay: listen EADDRINUSE: address already in use 0.0.0.0:8787

$ ss -tlnp | grep 8787
LISTEN 0 512 127.0.0.1:8787 0.0.0.0:* users:(("bun",pid=18638,fd=11))
```

`start` claims success; `status` contradicts it one command later; the relay's own log records the bind failure; and the socket belongs to an unrelated process.

`~/.agentbox/relay.pid` is also left holding a pid that does not exist, so the stale pidfile persists after the failed start.

## Why it matters

The relay dying is invisible at the point of failure. Downstream, in-box `agentbox-ctl` calls (`git push`, `checkpoint`, …) POST to the relay URL and reach whatever *else* is on that port — in my case an unrelated local HTTP service that answered with HTTP 502. The user sees an intermittent-looking 502 from a "running" relay, with no indication the relay never started. It took reading `relay.log` directly to find the actual `EADDRINUSE`.

A truthful failure here would have made this immediately self-diagnosing.

## Where it comes from

I traced this from the published bundle rather than a local build, so the exact call path is worth confirming — but the shape is clear: the success line is emitted from the intended endpoint URL rather than from a post-spawn liveness check.

Note `packages/relay/src/bin.ts:29` writes its `listening on …` line after `startRelayDaemon(...)` returns, and the daemon's own `EADDRINUSE` surfaces asynchronously to the log — so a spawn-and-report path that doesn't probe will report success regardless of whether the bind succeeded.

`ensureRelay()` in `packages/sandbox-docker/src/relay.ts` already has `/healthz` probe helpers (around `:530` / `:565`), so the machinery to verify is present in the same module.

## Suggested fix

[...]

PR #320, opened 2026-08-27
fix(relay): honor relay.port and report relay start truthfully

### LINKED ISSUE #301: relay.port is documented and settable but never read — host relay always binds 8787
## Summary

`relay.port` is declared in the config schema and documented as a supported key, but no code reads it. The host relay always binds the hardcoded `DEFAULT_RELAY_PORT` (8787), so a user whose 8787 is occupied has no supported way to move the relay.

Per CONTRIBUTING.md ("Stale public docs are treated as a bug"), filing this as a bug rather than a feature request: the key is documented today and does nothing.

## Version

- `@madarco/agentbox` 0.27.2 (relay commit `c9424ad0`)
- Linux, docker + e2b providers configured

## Reproduction

With any other process already listening on 8787:

```console
$ agentbox config set relay.port 8799 --global
$ agentbox config get relay.port --all
relay.port:
  effective: 8799   (global /home/adam/.agentbox/config.yaml)
  cli:       <unset>
  workspace: <unset>
  project:   <unset>
  global:    8799
  default:   8787

$ agentbox relay start
◇  relay running on http://127.0.0.1:8787

$ agentbox relay status
relay: not running

$ tail -1 ~/.agentbox/relay.log
agentbox-relay: listen EADDRINUSE: address already in use 0.0.0.0:8787
```

The configured 8799 is never bound or attempted; nothing ever listens on it.

## Where it comes from

The key is fully declared:

- `packages/config/src/types.ts` — `key: 'relay.port', type: 'int', description: 'Host relay TCP port (advanced).', advanced: true`
- `apps/web/content/docs/configuration.mdx` — `` | `relay.port` | int (advanced) | `8787` | host relay TCP port | ``

But the docker relay module ignores it — `packages/sandbox-docker/src/relay.ts:34`:

```ts
const PORT = DEFAULT_RELAY_PORT;
const ENDPOINT: RelayEndpoint = {
  url: `http://host.docker.internal:${String(PORT)}`,
  hostUrl: `http://127.0.0.1:${String(PORT)}`,
  port: PORT,
};
```

That module-level constant then drives the spawn (`relay.ts:223` — `[relayBin, 'serve', '--port', String(PORT), '--host', '0.0.0.0']`), the `/healthz` probes, and the box-facing URLs. Several call sites derive their own URL from the same constant rather than from config (`apps/cli/src/commands/{create,shell,codex,dashboard}.ts`, `packages/sandbox-docker/src/hub.ts`).

A `gh search code` for `relay.port` across the repo returns only the schema entry and the docs table row — no consumers.

The relay binary itself already supports the flag (`packages/relay/src/bin.ts:16`, `.option('--port <number>', …)`), so the daemon side is fine; the CLI just never passes the configured value through.

## Suggested fix

Resolve `relay.port` from config at the `ensureRelay()` spawn site and thread it through `RelayEndpoint` instead of the module-level constant, keeping `DEFAULT_RELAY_PORT` as the fallback. The endpoint is currently a module constant, so this likely means constructing it per-call.

Worth checking that the in-box URL and the `--add-host` mapping follow the same value — the code comments note the in-box relay port is deliberately distinct so a nested agentbox doesn't collide, so those two shouldn't be conflated.

Happy to send a PR if the approach sounds right.

## Context

Hit this because an unrelated local service already held 8787. The relay silently failed to start on every invocation, and in-box `agentbox-ctl checkpoint` calls reached the *other* service instead, which returned HTTP 502 — so the symptom presented as a flaky relay rather than a port conflict. The false-success message from `relay start` is filed separately.

madarco:
Fixes #301 and #302 — same code path, and they compound: the failure is silent, and the documented workaround did nothing.

> Stacked on #319 (the test-safety fix), which must merge first.

## #301 — `relay.port` was a dead key

Declared in the schema, documented in the config table, read by nothing. A repo-wide grep for `effective.relay.` only ever hit `controlPlaneUrl` and `custodyMax*Bytes`. The port lived as a compile-time constant in three independent places (`DEFAULT_RELAY_PORT`, `HUB_RELAY_PORT`, `AGENTBOX_HUB_PORT ?? '8787'`) plus string literals in the box env.

Everything now resolves through one place — `relayPort()` in `packages/sandbox-core/src/relay-port.ts`. `HUB_RELAY_PORT` is gone; the single remaining literal is that module's `FALLBACK_RELAY_PORT`. This closes the TODO the repo left itself at `docs/hub-api-single-path-plan.md:1328` ("if the relay port ever becomes configurable, thread it rather than duplicating").

Two decisions worth reviewing:

- **One key moves relay *and* hub.** The hub is the relay daemon plus the Next UI in one process and they are mutually exclusive on the port. Moving only the lean relay would leave the eviction/handoff logic — which probes exactly one port — silently broken. The load-bearing line is `AGENTBOX_HUB_PORT` in `spawnHub`'s child env: without it the hub keeps binding 8787 while everything else probes the configured port.
- **The env mirror is not a second config surface.** It is how children (the relay daemon, the hub server, the queue worker, the CLI entry the relay shells back into) inherit the port without re-reading layered config — same idiom as the existing `AGENTBOX_CLI_VERSION` stamp. Config is read once at startup by `applyRelayPortAtStartup`, mirroring `applyEngineOverrideAtStartup`.

Out of scope, deliberately: the Hetzner control-plane deploy port and `apps/hub/docker-compose.yml`. Those are a *remote* deployment's published port, persisted on its own `ControlPlaneDeployRecord.port`.

Also fixed while here: `AGENTBOX_HOST_RELAY_URL` was missing from `RUNTIME_ENV_BLOCKLIST`, so a checkpoint would have baked the creating machine's port into every restored box.

## #302 — `relay start` reported success on a dead relay

Two defects. The wait loop used `pingHealthz`, a bare 2xx check, so **any** process answering `/healthz` on the port passed as our relay — which is exactly how the reporter's unrelated service read as healthy while their relay was dead. And nothing watched the child, so a relay that exited instantly still cost the full timeout and left a dead pid in `relay.pid`.

Now: `fetchHealthz` (shape-validated body), an `exit` watcher, pidfile cleared on failure, the `relay.log` tail inlined, and the foreign occupant named with the way out. `ensureRelay` keeps its best-effort contract for `create`/`claude`; `relay start`/`restart` re-probe with `getRelayStatus()` before claiming anything, which also covers the "pid alive but /healthz silent" path. A hub serving the port is now reported as the hub — previously it said "relay running" when no relay process existed at all.

## Verified end-to-end, not just unit-tested

**#302**, against a real foreign occupant (`python3 -m http.server 8787`):

```
■  relay process exited (exit code 1) during startup
│  port 8787 is held by a process that is not an agentbox relay. Free it, or move the relay with
│  `agentbox config set relay.port <port> --global` (then `agentbox relay restart`).
│  --- last lines of /Users/marco/.agentbox/relay.log ---
│  agentbox-relay: listen EADDRINUSE: address already in use 0.0.0.0:8787
=== EXIT CODE: 1 ===
```

No `relay.pid` left behind; `relay status` agrees (`not running`).

**#301**, the escape hatch the reporter didn't have — with 8787 still occupied:

- `config set relay.port 8799 --global` → `relay start` → `relay running on http://127.0.0.1:8799`, confirmed by `lsof -iTCP:8799` and a real `/healthz`, not the exit code.
- `hub start` then takes 8799 (`ui:true`) — the leg that fails without the `AGENTBOX_HUB_PORT` fix.
- `relay start` while the hub holds it now prints `hub (serving the relay) running on http://127.0.0.1:8799`.
- `hub target --json` → `http://127.0.0.1:8799`, so the **macOS tray follows with no Swift change** (it resolves its base URL by shelling that command).
- A fresh docker box gets `AGENTBOX_HOST_RELAY_URL=http://host.docker.internal:8799`, reaches it, and a real relay-routed host action (`agentbox-ctl git fetch`) round-trips box → forwarder :8788 → host relay :8799 → host git. The in-box port stays 8788, so the nested-agentbox invariant holds.

Restored to the default port cleanly afterwards.

Gates: `pnpm build`, `pnpm typecheck`, `pnpm lint` green; **3906 tests pass**. New: `relay-port.test.ts` (resolution precedence, range/junk handling) and `relay-spawn-failure.test.ts` (early exit, log tail, foreign-occupant message, pidfile cleared).

## Docs

`configuration.mdx` (the `relay.port` row now says what it actually does), `hub.mdx`, `cli.mdx`, `docs/architecture.md`, the resolved TODO in `docs/hub-api-single-path-plan.md`, and CHANGELOG. The sibling tray repo's `CLAUDE.md` notes that `HubTarget.localFallback` still assumes 8787 for the CLI-unavailable case — behavior unchanged, no tray release needed.

https://claude.ai/code/session_01UHCTcn1S1s9a87iwcd9JfJ

---

> [!NOTE]
> **Medium Risk**
> Changes the default networking path for relay, hub, and new docker boxes when `relay.port` is set; startup/bind failures are now surfaced correctly but behavior on port conflicts shifts from silent misrouting to hard errors.
> 
> **Overview**
> **`relay.port` is wired end-to-end** through a shared `relayPort()` resolver in `@agentbox/sandbox-core`. The CLI pins it at startup (`applyRelayPortAtStartup`), hub spawn passes `AGENTBOX_HUB_PORT`, and docker boxes get `AGENTBOX_HOST_RELAY_URL` from the configured port. Port **8788** is rejected (in-box relay). Checkpoints no longer bake `AGENTBOX_HOST_RELAY_URL`.
> 
> **`agentbox relay start`/`restart` no longer claim success on a failed bind.** They re-probe with shape-validated `/healthz`, exit non-zero with log tail and foreign-process hints when the port is wrong, and distinguish hub vs lean relay in status output. `ensureRelay` spawn path got the same diagnostics (child exit watcher, `fetchHealthz` instead of bare 2xx, stale pidfile cleanup).
> 
> **Test harness hardening:** root `vitest.workspace.ts`, shared `scripts/test-home.ts` with `assertTempHome` / `resetTempAgentboxHome` so suites that wipe `~/.agentbox` cannot hit a real home.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit caa24a43c9e99616376bb5d528fdb23e298721f2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
## Live smoke test — docker + cloud + hub, all on a moved port

Re-verified end-to-end after a full provider reconfigure and rebake (all six `*-prepared.json` present).

| # | Check | Result |
|---|---|---|
| 1 | `relay start` while the hub holds the port | `hub (serving the relay) running on http://127.0.0.1:8787`; `relay status` → `running (served by the hub)` |
| 2 | `relay start` + `hub start` vs a foreign occupant on 8787 | both **exit 1** with the real cause (`EADDRINUSE` tail inlined), **no stale pidfile** |
| 3 | `relay.port=8799` → `hub start` | binds 8799 (confirmed by `lsof`), `ui:true`; 8787 still the foreign service |
| 4 | `hub target --json` | `local http://127.0.0.1:8799` — the tray follows with no Swift change |
| 5 | docker box created on the moved port | `AGENTBOX_HOST_RELAY_URL=…:8799`, registers with the hub, real `agentbox-ctl git fetch` round-trips |
| 6 | background `-i` job (queue worker → `submit.ts` enqueue poke) | job `done`, box created, claude tmux session **alive** |
| 7 | **e2b cloud box** on the moved port | registers (`boxes:3`); `agentbox-ctl git fetch` returns real refs from the host bundle — the poller drives fine on 8799 |
| 8 | restore to default | hub back on 8787, 8799 released, `hub target` follows |

Gates: `pnpm build`, `pnpm typecheck`, `pnpm lint` green; **3906 tests pass**.

## One real defect the smoke test surfaced — fixed in af8f37ab

`AGENTBOX_HOST_RELAY_URL` is baked in by `docker run -e`, and container env is immutable. So changing `relay.port` strands every **existing** docker box on the old port:

```
old port 8799 -> 000
new port 8787 -> 200
$ agentbox-ctl git fetch
agentbox-ctl git: relay returned 502
```

A bare 502 reads as a flaky relay rather than a port change — the same class of confusion #302 is about, and a user hits it precisely *because* they moved the port. `agentbox doctor` now names them:

```
[warn] box relay port  agentbox-qsmoke (:8799), agentbox-hubsmoke (:8799) still dial a relay port
                       this host no longer serves (relay.port is now 8787). Their host actions
                       (git push, cp, checkpoint) will fail with a 502 — the port is baked in at
                       create time, so recreate these boxes.
```

Verified it fires with stale boxes present and goes silent once they're recreated. Cloud boxes are unaffected (the host polls them, they never dial it). The docs caveat previously said such boxes "keep the port they were created with", which understated it — corrected to say they break, and to point at `doctor`.

Issue #260, opened 2026-07-26
wrapped-pty: status band corrupts on scroll over high-latency transports (sprites)


hexsprite:
Attaching to a **sprites** box renders correctly at first, then corrupts once content scrolls near the bottom: the status band is drawn twice at different rows, and band text ends up interleaved into a content line mid-screen.

## Not a sizing bug

Ruled out by measurement on a live box:

- terminal size propagates: host 143 cols produced tmux `pane=143x46` (one row for the band)
- SIGWINCH propagates: host resized 30x100 to 55x150, remote `stty size` reported `30x100` then `55x150`
- the transport does not split the stream across stdout/stderr: stderr carried **0 bytes** in TTY mode

## What is actually happening

`wrapped-pty/run.ts:969` restricts the OUTER terminal to rows `1..innerRows` so the inner program cannot scroll into the footer, and re-asserts it on relayout/resize (`:460`, `:789`, `:1095`).

But tmux inside the box emits its own `ESC[N;Mr` constantly. Captured from a real attach — 14 seconds, tmux running `seq 1 500`:

```
total bytes                       = 4601
ESC[r    (region reset)           = 0
ESC[N;Mr (region set)             = 57
ESC[?1049 (alt screen)            = 1
```

57 scroll-region writes in 14s, all landing on the same outer terminal as the wrapper band.

That is normally harmless: the inner PTY is sized to `innerRows`, so tmux computes regions inside that height and never addresses the band row. The problem is the **resize window**. When the alert band appears or disappears, `reservedRows()` changes, the wrapper moves its band and resizes the inner pty — and the remote repaints at the OLD size until the resize lands. On docker that window is microseconds. Over a network transport it is a round trip, and tmux paints into rows that are now band rows.

Hence: fine until something scrolls or the band changes, then band text stranded in content and a doubled band.

## Scope

The wrapper assumption — that an inner resize takes effect before the next paint — is shared, so hetzner/daytona/remote-docker are exposed too; sprites just has enough latency to hit it reliably. Reproduces with `agentbox claude --provider sprites` in a repo whose agent output scrolls.

## Possible directions

- Repaint/clear the reserved rows after a relayout rather than only redrawing the band, so a stale remote paint is overwritten rather than left behind.
- Defer the band move until the inner pty has produced output at the new size.
- Re-assert the scroll region on a short timer while a resize is in flight.

[...]

PR #321, opened 2026-08-27
feat(providers): declarative ProviderDescriptor — community providers in the hub, tray and CLI

### LINKED ISSUE #306: macOS tray/hub does not discover registered provider plugins
## Summary

A compatible provider plugin can be registered, loaded by the CLI, and used successfully, but it does not appear in the macOS tray app's **New Box → Provider** dropdown. The hub also rejects the plugin provider if a native client submits it directly.

This makes the provider-plugin surface CLI-only even though the tray obtains its provider choices from the hub API.

## Environment

- AgentBox CLI: `0.27.2`
- AgentBox macOS tray: `0.1.14`
- Provider SDK API: `v2`
- macOS, Apple Silicon

Observed with an `agentbox-provider-sprites` plugin, but the problem applies to any external provider plugin.

## Reproduction

1. Install/build a provider plugin that exports `providerModule` and declares a supported `providerApiVersion`.
2. Register it:

   ```bash
   agentbox plugin add .
   ```

3. Confirm the CLI sees it:

   ```bash
   agentbox plugin list
   agentbox doctor
   ```

   Both show the `sprites` provider, and this succeeds end-to-end:

   ```bash
   agentbox create --provider sprites --no-vnc
   ```

4. Open the macOS tray app and choose **New Box**.
5. Open the Provider dropdown.

## Expected

Registered, compatible provider plugins should:

- be returned by `GET /api/v1/providers`;
- appear in the tray provider picker;
- be accepted by `POST /api/v1/boxes` and lifecycle operations;
- expose their `readCredStatus`, optional freshness, and capability metadata through the same runtime provider-module contract used by the CLI.

## Actual

The tray requests:

```text
GET /api/v1/providers?freshness=1&hosts=expand
```

The embedded hub's `listProviders()` builds the response from the static built-in `PROVIDER_NAMES` list:

```js
return PROVIDER_NAMES.map((id) => { ... })
```

The create path is also built-in-only:

```js
if (!isProviderKind(provider)) {
  return { ok: false, error: `unknown provider ${provider}` }
}
```

Other hub paths similarly resolve providers through static `IMPORTERS` / `isProviderKind` checks. By contrast, the CLI and relay load `~/.agentbox/plugins.json` and resolve `providerModule` / `providerModules` correctly.

The remote-Docker entries visible in the tray are not generic plugin discovery; they are a special `hosts=expand` expansion for the built-in `remote-docker` provider.

Restarting the tray or hub cannot fix this because the inventory and mutation validation are static.

## Suggested direction

Introduce one runtime provider registry shared by the CLI and embedded hub:

1. Merge built-ins with compatible providers from `~/.agentbox/plugins.json`.
2. Use the plugin-capable provider-module loader for create and lifecycle mutations instead of gating on `isProviderKind()`.
3. Build provider-picker metadata from `ProviderModule`:
   - `provider.name`;
   - `readCredStatus()`;
   - optional `currentBaseFingerprintLive()`;
   - whether `prepare` is supported/required.
4. Do not assume every plugin needs a baked base image. A provider such as Sprites installs the runtime per box and has no `prepare` requirement.
5. Keep incompatible, missing, or failed plugins out of the picker with an actionable warning rather than crashing the hub.

## Regression coverage

- Register the sample external provider, then assert `GET /api/v1/providers` includes it.
- Assert `POST /api/v1/boxes` accepts the plugin provider and reaches its provider module.
- Assert incompatible and removed plugins are skipped safely.
- Assert plugin credential status is reflected without requiring built-in credential-key tables.
- Preserve `remote-docker` host expansion alongside plugin entries.

The public provider guide currently promises that plugins work with `agentbox --provider <name>`; extending the same registry to the hub/tray would make that provider surface consistent across clients.

madarco:
Closes #306.

A registered provider plugin was creatable from the CLI but invisible everywhere else: the hub's `listProviders()` built its response from the static `PROVIDER_NAMES` list, so a plugin never reached `GET /api/v1/providers` and therefore never reached the tray's **New Box → Provider** dropdown or the web create modal.

The deeper problem was that a plugin could declare only a name and an SDK major. Everything else a UI needs — a label, which credentials to prompt for, whether it needs a bake, whether it supports checkpoints/VNC/SSH — lived either in per-provider tables the hub and tray hardcoded, or in ~20 provider-name arrays across the CLI that a plugin could never enter.

## What this adds

One **`ProviderDescriptor`** that built-ins and plugins both declare: label, credential fields, bake story, and capabilities. Built-ins declare it in the `PROVIDERS` table; a plugin declares it on `ProviderModule.descriptor`, and `agentbox plugin add` snapshots it into `~/.agentbox/plugins.json` (v2). Every consumer — the CLI's sync `open --targets` probe, the hub's hot `listProviders` path, the tray — then resolves capabilities **synchronously and offline**, without importing an external package.

- `GET /api/v1/providers` lists built-ins and plugins alike, each row carrying its descriptor.
- The hub's readiness, credential and bake logic reads the descriptor; `PROVIDER_CRED_KEYS` is gone. The credentials/prepare routes accept plugin ids.
- The tray and the hub's settings card build their credential forms from `credentials.fields` and pace progress bars from `bake.*ProgressSteps`, instead of from hardcoded tables.
- `agentbox code`, `open`, `prune`, `checkpoint set-default` and `fork` gate on declared capabilities rather than name lists.

## Capabilities are declared, not derived

The original design inferred capabilities from optional-method presence. Probing the real modules showed that does not work: `createCloudProvider` defines `setInbound`, `enableDirectGit` and `checkpoint` on **every** cloud provider, so their presence signals which scaffold was used, not what the provider supports — and `sandbox-docker` has no `provider.checkpoint` at all despite `docker commit` checkpoints being the oldest feature in the repo.

The honest line is which object the method hangs off: `CloudBackend` methods are author-written, `Provider` methods are scaffold-supplied. So `prune`, `inbound` and `timeoutModel` are cross-checked against the backend by a guard test; everything else is declared.

## Nothing breaks for existing plugins

`descriptor` is optional and `SDK_API_VERSION` is unchanged, so a plugin built against an older SDK loads exactly as before. When one is missing, AgentBox derives what it can and defaults the rest to **reproduce pre-descriptor behavior** rather than to "safe = false" — `vnc`/`dind` stay `true` (the scaffold wires both; `false` would delete working UI), `pauseSemantics` stays `freeze`, and `bake.required` defaults **false** so a create that works today is never newly blocked. Loading a plugin module back-fills the derived descriptor, so an existing `plugins.json` migrates with no user action. Pinned by `provider-descriptor.test.ts`.

## Verified against a real third-party plugin

Live-tested with the published `agentbox-provider-sprites` (SDK v2, declares no descriptor) — the plugin from #306:

- Lists as `configured`, creatable, nothing blocked.
- Derivation gets `prune: true` and `timeoutModel: "inactivity"` right from its backend, the latter matching Sprites actually hibernating when idle.
- Its credential status comes from its own `readCredStatus()` (`label: "sprite CLI"`), not a built-in key table — #306's "without requiring built-in credential-key tables".
- `agentbox prune --provider sprites --dry-run` enumerated real orphan sandboxes; that command was rejected outright before.
- The tray's own `Codable` decodes all rows including the plugin.

Two things only a *declared* descriptor can fix for Sprites, worth an upstream PR to `philippgerard/agentbox-provider-sprites`: the fallback guesses an `apiKey` credential field (it wants `SPRITE_TOKEN`), and it inherits `vnc: true` though Sprites ships no VNC stack. Neither is a regression — both match pre-descriptor behavior.

## Behavior changes worth calling out

- **DigitalOcean gains `agentbox code` / `open`.** It is a VPS with the same per-box identity file as Hetzner; the old arrays omitted it while listing it in the direct-box-ssh set.
- **`checkpoint set-default --provider <plugin>` now warns** that a plugin has no per-provider config key, so the write lands on the cross-provider default docker boxes also read. Previously that would have been silent.
- `pauseSemantics: 'freeze' | 'stop'` instead of a boolean: Hetzner and DO are literally `pause ≡ stop`, and a boolean would have hidden a pause control that usefully stops billing. It is a labelling hint, never a reason to hide.

## Notes for the reviewer

- The tray half is a separate commit in `madarco/agentbox-tray` (`3db893a`, on `nightly`).
- `@madarco/agentbox-provider-sdk` is bumped to `2.8.0` here but **not published** — that stays a manual step.
- Plan and rationale: `docs/provider-descriptor-plan.md`.

https://claude.ai/code/session_01SKLmAbYKtGVKNzKG3j9eqp

---

> [!NOTE]
> **Medium Risk**
> Touches shared relay/hub port binding, plugin registry snapshots, and hub VNC redirect auth—areas where misconfiguration or stale state affects every box, though changes are heavily tested and mostly fail-closed with clearer errors.
> 
> **Overview**
> **Community provider plugins** now surface in create flows and credential UIs via a shared **`ProviderDescriptor`** (built-ins in config, plugins on `ProviderModule.descriptor` and snapshotted in `plugins.json`). The CLI stops using hardcoded provider name lists for **`code`**, **`open`**, **`prune`**, **`checkpoint set-default`**, and **`fork`**, and instead gates on declared capabilities (e.g. DigitalOcean gets SSH/`open` behavior it was missing before).
> 
> **Relay and hub behavior** is tightened: **`relay.port`** is applied at CLI startup and threaded through loopback URLs; **`agentbox doctor`** warns when docker boxes still dial an old baked-in relay URL; **`relay start`** re-probes `/healthz` and fails loudly on a bad bind. Hub/tray-oriented fixes from the changelog (hub vs relay on 8787, stale bundle warning) ride along in the same release notes.
> 
> **VNC access** for expiring cloud URLs moves to **mint-at-click**: new hub routes **`GET /api/v1/boxes/:id/vnc`** and **`/boxes/:id/vnc`**, **`agentbox screen`** prefers the owning hub, and **`agentbox list`** builds `(VNC)` links through the hub when the payload has no static `vncUrl`.
> 
> Smaller but user-visible changes: **GitHub Enterprise / SSH `Host` aliases** for hub git token and direct-push creds; **size ignored** notes on `config set` and queued `-i` creates; **unknown subcommand** errors on groups with default actions; **wrapped attach chrome** gated on escape-sequence boundaries (fixes garbled footers on fine-grained PTY transports).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1616940b4574a7bdace832004b806e150e0dcab. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -11,6 +11,16 @@ CLI, not the raw commits.
 
 ## [Unreleased]
 
+### Added
+
+- Community providers now appear in the create pickers (web UI and tray) and get
+  a real credential form, from a new `ProviderDescriptor` a plugin declares and
+  `agentbox plugin add` snapshots. Existing plugins keep working unchanged —
+  declaring one is optional.
+- `agentbox code`, `open`, `prune`, `checkpoint set-default` and `fork` now gate
+  on a provider's declared capabilities instead of hardcoded name lists, so a
+  plugin that supports them gets them. DigitalOcean gains `code`/`open` this way.
+
 ### Fixed
 
 - `relay.port` now actually moves the host daemon. It was documented and
--- a/apps/cli/src/commands/_open-in.ts
+++ b/apps/cli/src/commands/_open-in.ts
@@ -11,6 +11,7 @@
 import { existsSync as realExistsSync } from 'node:fs';
 import { homedir as realHomedir } from 'node:os';
 import { delimiter, join } from 'node:path';
+import { listProviderDescriptors } from '@agentbox/sandbox-core';
 
 // `finder` is a first-class open target: `agentbox open <box>` (no `--in`) is
 // equivalent to `--in finder` and sshfs-mounts /workspace + reveals it. It's an
@@ -30,6 +31,22 @@ export const OPEN_IN_APPS: readonly OpenInApp[] = [
 ];
 
 /**
+ * Providers matching a capability, from the descriptor table — built-ins AND
+ * registered plugins. Sync + offline (the descriptor is a snapshot, never a
+ * module load), which is what `detectOpenTargets` needs: it must stay a cheap
+ * fs-only probe.
+ *
+ * These replaced hardcoded name arrays, so a community provider that declares
+ * the capability is listed here automatically instead of being permanently
+ * opted out.
+ */
+function providersWith(cap: 'persistentSsh' | 'ssh'): string[] {
+  return listProviderDescriptors()
+    .filter((d) => d.capabilities[cap])
+    .map((d) => d.name);
+}
+
+/**
  * Providers whose per-box SSH identity outlives the CLI call (see
  * packages/sandbox-core/src/cloud-ssh.ts). Codex connects on its own later, so
  * an expiring token credential (Daytona) or no SSH at all (vercel/e2b)
@@ -37,35 +54,21 @@ export const OPEN_IN_APPS: readonly OpenInApp[] = [
  * per-box key that persists under the box dir (only the loopback host port
  * changes across restart, and `agentbox open`/start re-syncs `~/.ssh/config`).
  */
-export const PERSISTENT_SSH_PROVIDERS: readonly string[] = [
-  'docker',
-  'hetzner',
-  // remote-docker qualifies for the same reason docker does: a per-box key that
-  // outlives the CLI call. Its box sshd is reached by `ProxyJump`ing through the
-  // engine (ssh's own transport — no AgentBox-held tunnel to keep alive), and
-  // start/open re-resolve the published port into `~/.ssh/config`.
-  'remote-docker',
-];
+export function persistentSshProviders(): string[] {
+  return providersWith('persistentSsh');
+}
 
 /**
- * Providers `agentbox code` can attach an IDE to: docker via the Dev Containers
- * attached-container URI, clouds via a Remote-SSH alias.
+ * Providers `agentbox code` can attach an IDE to (docker via the Dev Containers
+ * attached-container URI, clouds via a Remote-SSH alias) and `agentbox open` can
+ * sshfs-mount `/workspace` from. Both need the same thing — real SSH into the
+ * box — so both read `capabilities.ssh`. Vercel/E2B have none (their
+ * `buildAttach` yields an SDK PTY bridge), so `open` fails fast with a readable
+ * pointer to `agentbox download`.
  */
-export const IDE_PROVIDERS: readonly string[] = ['docker', 'hetzner', 'daytona', 'remote-docker'];
-
-/**
- * Providers `agentbox open` can sshfs-mount `/workspace` from — they expose real
- * SSH: docker's localhost sshd, Hetzner's VPS, Daytona's token gateway. Vercel/E2B
- * have no SSH (their `buildAttach` yields a non-SSH `sbx exec` / SDK PTY bridge),
- * so `open` fails fast with a readable pointer to `agentbox download`. Plugin
- * providers stay opted out until they declare SSH support.
- */
-export const SSH_MOUNT_PROVIDERS: readonly string[] = [
-  'docker',
-  'hetzner',
-  'daytona',
-  'remote-docker',
-];
+export function sshProviders(): string[] {
+  return providersWith('ssh');
+}
 
 export interface DetectSeams {
   env: NodeJS.ProcessEnv;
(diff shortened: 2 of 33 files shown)


madarco:
Retargeted the base from `main` to `nightly`: this branch is cut from `origin/nightly`, so the diff against `main` was carrying 18 unrelated nightly commits (150 files). The two findings above are in `screen.ts` and `escape-boundary.ts`, both from that pre-existing nightly work rather than this change — they should be raised against whichever PR introduced them.

The diff is now the 7 commits of this change (35 files).

bugbot run

madarco:
Force-pushed: my last commit had accidentally swept in two unrelated draft docs (`docs/agent-catalog-plan.md`, `docs/openclaw-hosting-plan.md`) that were untracked in the working tree. They're out of the branch and restored untracked locally, contents unchanged. Diff is now 33 files.

For the record, the two findings above were reviewed against the old `main` base and are not from this change — `apps/cli/src/commands/screen.ts` comes from `ccd5d821` ("feat(vnc): surface the box desktop on cloud providers") and `apps/cli/src/wrapped-pty/escape-boundary.ts` from `89971e55` (#260). Neither file is touched by this PR.

bugbot run

PR #323, opened 2026-08-27
feat(hub): make box cp reach the machine that holds the files

madarco:
## The bug

With a remote hub, an in-box `agentbox-ctl cp` never ran at all.

Every cloud box resolves `topology: 'control-plane'` and forwards `/rpc` to the control box — which
holds no checkout of your project. For a box it created it records `workspacePath:
/tmp/agentbox-hub-worker-<jobId>`, the create job's temp clone, and the worker deletes it. The relay
passed that path as the **cwd** of the CLI it spawns, and Node reports a missing cwd as `spawn <bin>
ENOENT`. Measured on a live control box: exit 127 for both directions, after the gate had
auto-approved the copy as "contained" against that same phantom directory.

The shipped docs promised the opposite (`deployed-hub.mdx`: "cp … still needs your laptop up"), so
this was a defect, not a documented gap.

## The change

The control box **brokers** instead of executing. It parks `cp.*` for the machine that owns the box;
that machine's relay long-polls `/admin/hostreach/*`, gates the copy with its own prompts, and runs
it against the real files. The gate moves with the executor — deciding containment on the broker
means judging a path that doesn't exist there.

With that machine offline, neither direction fails blind:

- **`fromHost`** is served from a **custody cache** a previous copy wrote — its own approval (parked
  on the hub, answerable from the web UI/tray) and labelled with the entry's age, so an agent can
  tell a cached file from a live one. Nothing cached → an error naming *both* facts.
- **`toHost`** has nowhere to land, so the hub pulls the bytes out of the box and **parks** them;
  your machine lands them behind a confirm on its next connect. The box gets exit 75 — accepted, not
  delivered — never a false success.
- `agentbox cp <file> hub:` writes the same cache entry by hand, keyed so a box's own
  `cp fromHost ./data.csv` finds it.

`relay.hostReachTimeoutMs` (60s) tunes how long the hub waits before falling back.

## Live verification (real control box, built from this branch)

| path | result |
| --- | --- |
| live `fromHost` | file created on the Mac *after* the box reaches it; approval on the Mac |
| live `toHost` | lands on the Mac at the requested path |
| cached `fromHost`, relay stopped | approval parks on the hub; box gets the file + the cache note |
| cold miss | non-zero exit naming the offline machine *and* the empty cache |
| `cp <file> hub:` → box read, offline | file lands in the box, marked cached |
| parked `toHost` | exit 75, item in the outbox |
| outbox drain on reconnect | prompt on the Mac, file landed after approval |

## Bugs live testing caught that unit tests didn't

- **`hub expose` would have broken.** There the laptop *is* the control box and the files really are
  where the record says. Keyed on whether the workspace exists rather than on the hub's profile.
- **The outbox drain blocked the poll loop** while its confirm was open, so the hub declared the
  machine gone and live copies failed over to the cache with the user sitting right there.
- **A copy handed to a relay that then restarted was never re-offered** — the went-away sweep
  couldn't see it, because the replacement poller kept the machine "reachable". Pollers now identify
  themselves.
- **`/admin/hostreach` inherited the loopback bypass** (caught by my own routing test): on a control
  box every request arrives via Caddy and looks loopback, so a box could have read other boxes'
  parked actions and forged their results. Admin bearer only, like custody.

## Known caveat

One case failed once — a `hub:`-uploaded entry read from a box delivered nothing while the hub
logged `served` — and is not reproducible; it passes now. Rather than claim a repair I didn't make,
the serve no longer trusts the inner CLI's exit code: it requires the CLI's own `copied to` line,
checks the staged sources exist, and logs what the copy actually said when they disagree. Ruled out:
the stored entry, the command shape (hand-run in the hub container works), and a workspace re-sync.
Written up in `docs/plans/box-cp-host-reach-plan.md` and the backlog.

Also inherent: an offline copy takes ≥60s plus your approval time, so an agent wrapping
`agentbox-ctl cp` in a short `timeout` will see its client give up while the copy still completes.

## Not in scope

`download.workspace` / `download.env` / `download.claude` have the identical defect. The channel is
built so each is a routing line; the docs say plainly that only `cp` is fixed. Backlog also carries
cache eviction and the `~`-expands-to-the-box's-home problem.

Full suite green (1362 + 366 + 180 + 137), typecheck and lint clean.

---

> [!NOTE]
> **High Risk**
> Changes authentication, custody, and cross-machine copy brokering on the control plane—security-sensitive paths where a false success or wrong machine executing a copy would be severe. Scope is large but focused on fixing a broken remote-hub cp path with live verification.
> 
> **Overview**
> Fixes **in-box `cp` with a remote control box**, where copies used to fail because the hub had no real project checkout (spawn against a deleted create-job temp dir → `ENOENT`).
> 
> The control box now **brokers** `cp.toHost` / `cp.fromHost` instead of running them locally: actions park on a **`HostReachQueue`**, and the laptop’s relay **long-polls** `/admin/hostreach/*` (admin bearer, not loopback), prompts on the real project, runs the copy, and posts results. **`ensureHostReachDrainer`** / startup env wiring starts that poller when you attach, adopt, or use hub boxes without a local-only create path.
> 
> **Offline behavior** is explicit: **`fromHost`** can be served from a **custody cp cache** (approval on the hub, stdout notes cache age); **`toHost`** is **parked in an outbox** (exit 75) and **drained** when the relay reconnects. Successful live reads also **warm the cache**; **`agentbox cp <paths…> hub:`** pre-loads the same keys. Cache keys use the **request path** (`rel:` / `abs:`), not each side’s resolved absolute path.
> 
> Adds **`relay.hostReachTimeoutMs`** (hub reads config/env), **`hub status`** nudge when the local relay isn’t draining, and stricter **cached serve** validation (requires CLI `copied to`, staged file checks). Docs, changelog, plan, and tests cover the flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6287f380bbec52093609c45daadb6b96565d89aa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #324, opened 2026-08-28
feat(hub): run host tools where they were granted, and stream host actions

madarco:
Follow-on to #323. Same channel, the two gaps it left — plus the transport question.

## 1. Host tools were broken on a control box, exactly like `cp` was

`tool.*` was never brokered to the user's machine (only `cp.*` was), so on a hub box it ran on the
control box — and `runToolRpc` resolved grants against `lookupCloudBox().workspacePath`, the create
job's temp clone that the worker deletes. So `tool.list` walked a directory that no longer exists
and quietly returned what was left, and `tool.run` would have run **the control box's binaries**
under the user's tool name. `agentbox tools add jq` on your Mac had no effect on what a hub box saw.

`tool.*` now travels the same path as `cp`: the grant is a file on the machine that made it, and the
binary it names is installed there. **No fallback** — with that machine off the box is told plainly
and nothing runs, because a same-named binary elsewhere is a different program, not an older copy of
the same one.

## 2. The box no longer polls for grants

`tool-links-watcher.ts` reconciled every 60s because "the RPC is the one channel every provider
already has". But `Provider.exec` is a required interface method, so the host can just say it:
`agentbox tools add/rm` now runs `agentbox-ctl tool relink` in each running box and reports which
ones took it. The box keeps a single reconcile at daemon start — precisely the case a push cannot
cover (paused when the grant changed, or created after it).

## 3. SSE for host-reach, long poll as the fallback

The channel was already a long poll with instant hand-off, so this is not a latency fix — the gain
is that presence becomes a live connection instead of heartbeat bookkeeping. The poll stays as the
fallback for proxies that buffer event streams and for control boxes too old to serve the route;
neither is detectable without trying, so the client tries the stream, falls back on the first hard
evidence against it (404, non-stream content type, or nothing at all within 10s), and retries the
stream every 5 minutes. Subscribers reuse the queue's existing waiter path, so the ownership,
decline and orphan rules from #323 have one implementation, not two.

## Live verification (control box `46.225.235.16`, built from this branch)

- `agentbox tools add jq` → `applied in toolbox` in ~1s; the box's `tool list` shows `jq`
  immediately.
- `jq --version` in the box → `jq-1.8.1`, matching **the Mac's** jq; the Mac's relay logs
  `executing tool.run`, the control box only logs `rpc … method=tool.run`.
- `agentbox relay stop` → the same call exits 69 after ~60s naming the machine that granted the
  tools. Nothing runs on the control box.
- `host-reach: streaming host actions (event stream)` on the Mac, `machine 7de35e0a streaming` on
  the hub — SSE works through Caddy.
- `cp` regression over the new transport: `fromHost` 2s with correct content, `toHost` landed on the
  Mac at the requested path.

## Found by testing

- **`stop()` could not interrupt an open stream** — the unit test hung on it. `agentbox relay stop`
  would have waited for the control box to drop a connection it has no reason to drop.
- **A box created before this change** answers `unknown command 'relink'`. It is neither broken nor
  stale (its older ctl still polls), so it is now reported that way instead of as unreachable.

The SSE fallback paths are covered by unit tests against real sockets (404, 200-with-wrong-type);
the live run exercised the stream itself, not a buffering proxy.

Full suite green (1365 + 562 + 366 + 180 + 137), typecheck and lint clean.

---

> [!NOTE]
> **Medium Risk**
> Changes where hub boxes execute host tools and how grants sync across machines; mistakes could break remote workflows or mis-report box state, though behavior is gated and heavily tested with poll fallback.
> 
> **Overview**
> Fixes **host tools on a deployed hub** by routing `tool.*` through the same host-reach channel as `cp`, so execution and grant files stay on the machine that granted them—not on the control box’s deleted create clone. When that machine is offline, boxes get a clear error with **no substitute binary** (unlike cached `cp` reads).
> 
> **Grant propagation** stops relying on a 60s in-box poll: `agentbox tools add/rm` pushes `agentbox-ctl tool relink` to running boxes via `Provider.exec`, logs which boxes applied, and distinguishes legacy boxes that still poll from unreachable ones. In-box `ToolLinksWatcher` only reconciles at startup (plus explicit relink).
> 
> **Host-reach transport** adds `GET /admin/hostreach/events` (SSE) with shared `SseFrameReader` and queue **subscribers**; the PC relay prefers the stream, falls back to long poll on 404/wrong content-type/10s silence, and retries SSE periodically. `relay stop` can tear down an active stream so shutdown doesn’t hang.
> 
> Docs/changelog and tests cover routing, stream fallback, relink edge cases (empty revoke list, failed grant read), and push reporting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 335cc9a7522fbd9ca0884539d06dbadd2d1dcbf2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #327, opened 2026-08-28
fix(box): pin Claude Code's classic renderer inside boxes

madarco:
Claude Code's fullscreen renderer (alternate screen + virtualised scrollback) repaints **differentially** — it skips cells it believes are already blank. Over a network transport that assumption breaks down and stale characters are left behind.

They only ever show where the screen is otherwise blank, which is why the symptom reads as garbled text in the *gaps*. The two-space indent of an agent reply is the widest gap, so it looks like the first two columns are corrupted:

```
Ap
DaOther top-level dirs
 t- docs/ — domain and migration notes
Da  glossary, report specs)
Th- design/ — standalone HTML/JSX design mockups
```

Nothing repaints those cells afterwards, so the debris survives until a full redraw — hence "resizing the terminal fixes it".

## Not ours

Isolated by attaching to the same box three ways: `agentbox shell` is clean, `claude` with `/tui default` is clean, `claude` with the fullscreen renderer garbles. Worth reporting upstream separately.

## The fix

New `box.claudeTui` — `default` | `fullscreen` | `auto`, defaulting to `default`. Boxes pin the classic renderer until the upstream repaint is fixed; `agentbox config set box.claudeTui auto` reverts.

It maps to Claude Code's own overrides, both read off the shipped binary (v2.1.250) rather than guessed:

| mode | env |
| --- | --- |
| `default` | `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` |
| `fullscreen` | `CLAUDE_CODE_NO_FLICKER=1` |
| `auto` | *(nothing — Claude decides)* |

These outrank the `tui` key in `~/.claude/settings.json`, so it holds whatever a box's Claude config volume happens to carry.

## Why `/etc/agentbox/box.env`

It goes in the box env file rather than only onto the agent's launch command: `/etc/profile.d/agentbox.sh` sources that file in every login shell, so a `claude` started by hand from `agentbox shell` gets the same renderer as the session AgentBox starts.

- **docker / remote-docker** — `create.ts`'s `agentboxEnv`, which feeds both `docker run -e` and box.env.
- **cloud (e2b / vercel / hetzner / daytona / DO)** — `buildBootstrapEnv`, re-read on every kick, so flipping the key takes effect on the box's next start without recreating it.

Existing boxes need a stop/start to pick it up.

## Tests

- `packages/core/test/claude-tui.test.ts` — the mode → env mapping, including that the two contradictory variables are never both set.
- `packages/sandbox-cloud/test/bootstrap-env.test.ts` — the renderer reaches **both** the process env and box.env, honours an explicit `fullscreen`, and sets nothing for `auto`.

The repo's `schema-drift` guard caught a missing `user-config.schema.json` entry; added.

`pnpm typecheck` 29/29 and `pnpm test` 28/28 green.

---

> [!NOTE]
> **Low Risk**
> Terminal UX and launch-env wiring only; no auth, data, or network policy changes. Broad touch across Claude start paths but covered by new unit tests.
> 
> **Overview**
> Fixes garbled/stale characters when attaching to **Claude Code** in a box over a remote terminal by defaulting boxes to the **classic** renderer instead of Claude’s differential **fullscreen** repaint.
> 
> Adds **`box.claudeTui`** (`default` | `fullscreen` | `auto`, default `default`) and **`claudeTuiEnv`** in `@agentbox/core`, mapping modes to Claude’s own env overrides (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN` / `CLAUDE_CODE_NO_FLICKER`).
> 
> The setting is applied on **every agent start path**, not only `/etc/agentbox/box.env`: **Docker** forwards both TUI vars via new **`claudeSessionEnvFlags`** on `docker exec -e` in `startClaudeSession` (tmux runs `claude` without a login shell); **cloud** (and shared detached launch) injects `export …` before `exec` in **`buildCloudAttachInnerCommand`**. Call sites pass **`workspacePath`** / **`claudeTui`** so project config resolves correctly; config schema, changelog, and docs describe the option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6367cd192cf1ee92f87143d5da42792e58da1030. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #328, opened 2026-08-29
feat(agents): one agent per box — data-driven agent catalog + derived image layers

madarco:
Makes a box carry **only the agent it was launched for** — binary and credentials — and turns "add an agent" into a row of data rather than a Dockerfile edit.

Targets `feat/custom_agents`; the cloud half follows in a second PR.

## Why

Every box got all three agent binaries, and `create.ts` mounted a config volume for each agent whose host directory merely *existed* — so `agentbox claude` produced a box carrying codex and opencode credentials it would never use, on top of a 4.75 GB image.

## What changed

**Agents are data.** `AGENT_SYNC_SPECS` gains `binary` + an `install` recipe (npm / script / exec, `runAs`, apt deps, `postInstall`). One catalog now drives three consumers — the derived image layer, the cloud provision scripts, and `ensureAgentInstalled` against a live box — so a baked agent and a runtime-added one are installed identically.

**One `ensureAgentInstalled`** replaces the near-identical `ensureCodexInstalled`/`ensureOpencodeInstalled` twins and runs over `SyncTransport`, so **cloud boxes gain on-demand install for the first time** (previously a cloud box booted from a snapshot predating an agent had no path to it at all). Claude gains an installer it never had.

**The base image is agentless**, and each agent is added as a thin `FROM base` layer:

| | before | after |
|---|---|---|
| base | 4.80 GB | **3.07 GB** (100% shared) |
| `dev-claude` unique | 1.279 GB | **434 MB** |
| `dev-codex` unique | 1.518 GB | **702 MB** |

A *full* variant build would have to declare the agent ARG above Playwright/Chromium/VNC, so every one of those layers diverges per agent. Deriving avoids that and lets CI publish **one** image instead of one per agent set.

**Per-variant prepared records.** The single `base` slot in `docker-prepared.json` meant building codex overwrote the note that claude was prepared, so alternating agents rebuilt every time. Each variant now keeps its own record; `base` is still stamped so `prepare --status`, the freshness nag and custody adoption are unaffected.

**Dockerfile layer reorder** (first commit, independently useful): `ctl/bin.cjs` inlines core/sandbox-core/relay, and its COPY sat above every expensive layer — so a one-line TypeScript change invalidated ~1.8 GB. Moving the cheap COPYs to the end means a ctl edit now rebuilds 16 layers in 1.7s.

## Verified live, not just green tests

- **claude-only box**: `agentbox/box:dev-claude`, claude runs 2.1.236, codex/opencode absent, `~/.codex` and `~/.local/share/opencode` absent, zero codex/opencode mounts, real authenticated `claude -p` turn.
- **codex-only box**: codex 0.150.1, `bwrap` present (the gated apt dep works), only `agentbox-codex-config` mounted.
- **runtime add**: `agentbox claude start <codex-box>` installs claude and completes a **real authenticated turn**; codex's own auth is untouched and both tmux sessions coexist.
- **no rebuild thrash**: base 56s → +claude 23s → +codex 21s → claude again 0s "image up to date".
- Restoring the old single-slot prepared record reproduces the rebuild, so that fix is load-bearing.

Two bugs were found *only* by installing for real, which is why the live runs are listed above:
1. The Claude installer is a bash script and was being run with `sh` (dash on Ubuntu) — `Syntax error: "(" unexpected`. The transport-level unit tests assert argv, not that the box can run it.
2. A runtime-added agent came up **unauthenticated with no error**. Docker fixes mounts at `docker run`, so its config volume can never be attached to a running container; the credential is now pushed as a file, which is what every cloud provider already does per create.

## Not in this PR

Cloud derived snapshots (hetzner/vercel/e2b), `--agents` through `PrepareOptions` → hub API → queue → worker, cloud-side selection (`stageAllAgentStatic` still bakes all three), checkpoints recording their agent set, and publishing the agentless base + CI matrix — until that lands, the first build on a cold machine is local rather than a pull.

**Cloud impact — corrected.** Hetzner, DigitalOcean, Vercel and E2B are unaffected: they bake from their own provision scripts, which still install all three agents, so `ensureAgentInstalled` is a no-op there.

**Daytona is the exception.** It bakes from `Dockerfile.box` (container class) or from that same image published to GHCR (linux-vm), and this PR makes that Dockerfile agentless — so a *fresh* `prepare --provider daytona` now yields a base with no agents and falls back to per-box `ensureAgentInstalled` (~30-90s on first use). It degrades rather than breaks, and existing daytona bases are untouched until re-prepared. Daytona gets derived snapshots in the cloud follow-up.

Status table: `docs/agent-catalog-plan.md`.

---

> [!NOTE]
> **Medium Risk**
> Changes core docker image layering, box create/mount selection, and credential seeding for on-demand agent installs; mistakes could break login, wrong mounts, or stale image cache behavior, but cloud paths are explicitly unchanged.
> 
> **Overview**
> Docker boxes now ship **only the agent you launch** — binary, config volume, and credentials — instead of baking all three agents and mounting every agent directory the host happens to have.
> 
> **Image model:** `Dockerfile.box` is **agentless** (~3 GB shared base). `ensureImage` builds or pulls the base, then adds a thin `FROM base` layer per agent set (`agentbox/box:dev-claude`, etc.) using the same `install` recipes in `AGENT_SYNC_SPECS`. `variantFingerprint` and per-variant entries in `docker-prepared.json` stop alternating agents from rebuilding every time. Frequently changing COPYs (e.g. `agentbox-ctl`) were moved to the end of the Dockerfile so ctl edits no longer invalidate Playwright/VNC layers.
> 
> **Runtime:** Generic `ensureAgentInstalled` over `SyncTransport` replaces duplicated codex/opencode installers and adds **`ensureClaudeInstalled`** for dashboard/switch flows. Adding a second agent to a running box installs the binary and **pushes credentials as a file** (mounts cannot be added post-`docker run`). CLI create/login/queue paths pass `agents: [...]` and request the derived image ref where login containers need the agent binary.
> 
> **Docs:** New `docs/agents.md`, updates to `CLAUDE.md`, `run-an-agent.mdx`, and `agent-catalog-plan.md` status. Cloud provision scripts still install all three agents — no cloud regression in this PR.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1297d7448f4ab2eabaf8ebf7aa05d7ef972d044. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #329, opened 2026-08-29
Cloud: per-agent credential isolation + agentless hetzner base with derived snapshots

madarco:
The cloud half of the agent catalog. #328 made the docker base agentless with a
thin `FROM base` layer per agent; this does the equivalent for cloud, plus the
credential isolation that had to come first.

## Isolation (3 commits)

A cloud box carried every agent's credentials regardless of which agent it was
launched for. `ensureAgentVolumesForCloud`, the credential reconcile and the
home-dir ownership pass all take an agent filter now, `CreateBoxRequest.agents`
and `BoxRecord.agents` carry the selection (durable, so resume knows it), and
`planPropagateTargets` no longer pushes a credential into every running cloud
box with no agent check.

## Agentless hetzner base + derived snapshots (2 commits)

`install-box.sh` installs no agents; the per-agent home and credential dirs move
onto each agent's `install.postInstall`, so a baked agent and one added later by
`ensureAgentInstalled` produce the same layout. `prepare --agents claude` boots
the base, runs just that recipe over ssh and re-snapshots.

Prepared state gains a `variants` map (schema 3, lossless migration from 2) so
baking codex doesn't invalidate the claude snapshot, and each bake reaps the
snapshot it supersedes — only its own variant, only after the replacement is
recorded. Hetzner had no base-snapshot GC at all before this.

### Three things only the live run found

1. **Nothing on the cloud path called `ensureAgentInstalled`.** A box on an
   agentless base launched a missing binary; the only symptom was
   `no server running on /tmp/tmux-1000/default` at attach. Cloud create now
   probes and installs before seeding credentials — the order matters, since
   postInstall creates the `~/.agentbox-creds/<agent>/` the credential lands on.
   This applies to every cloud provider, not just hetzner.
2. **The base pin defeated variant selection.** Every bake pins
   `box.imageHetzner` to the base's description, so at create time the ref named
   our own base and `resolveImageId` never reached the variants map.
3. **`state.base` must stay the agentless base.** Provider-generic readers
   (freshness, bake sharing, custody adoption) read `base.contextSha256` and
   assume it describes the agentless context; pointing it at the newest bake
   made `doctor` report a permanent false stale.

A variant bake also skips base adoption, the `box.image<Provider>` pin and
custody sharing — all three are single-slot and would otherwise make every box
on the provider boot one agent's snapshot.

## Verified live on hetzner

| check | result |
|---|---|
| agentless base | 1.87 GB / ~6 min (was 2.61 GB / 10-15 min) |
| box on the base | installs claude at create, answers a real `claude -p` turn |
| `--agents claude` snapshot | ~3.5 min; box boots with claude present, no install |
| isolation | only `~/.agentbox-creds/claude`; holds across pause/resume |
| no thrash | codex bake left the claude record and snapshot untouched |
| GC | forced codex re-bake reaped only its own predecessor |
| skip-fast | named the right variant |

Test boxes destroyed; no orphan servers left.

## Notes

- Vercel, E2B and DigitalOcean still bake all three agents, so their behaviour
  is unchanged. Daytona bakes from the now-agentless `Dockerfile.box`, so it
  falls back to the per-box install — that is the next PR.
- `--agents` had never been documented publicly; `cli.mdx` and `hetzner.mdx`
  cover it now, alongside `docs/agents.md`.

---

> [!NOTE]
> **Medium Risk**
> Changes credential mounting, resume reconcile, and snapshot selection across cloud create/prepare; behavior is backward-compatible when `agents` is omitted but mis-wiring could leak or drop credentials for migrated callers.
> 
> **Overview**
> This PR adds **per-agent cloud boxes**: create paths pass an authoritative `agents` list so only that agent gets credential mounts, forwarded API keys, seeding, and resume-time reconcile. The choice is stored on **`BoxRecord.agents`**, and credential fan-out skips boxes that were not created for that agent.
> 
> **`agentbox prepare --agents`** (CLI, hub API, queue worker) bakes optional **derived snapshots** on top of an **agentless Hetzner base**. `install-box.sh` no longer installs agents or credential pivots; those live on each agent’s `install.postInstall`. Variant bakes boot the base over SSH as `vscode`, run registry recipes, and re-snapshot. **`hetzner-prepared.json` schema 3** adds a `variants` map (per agent set), keeps `base` agentless-only, reaps superseded snapshots, and **skips** base adoption, global `box.image*` pin, and custody share for variant bakes. **Create** picks the matching variant in `resolveImageId` even when the config pin names the agentless base.
> 
> Cloud create now runs **`ensureAgentsInstalledForCloud`** before credential seeding so agentless bases do not fail silently at attach. Claude’s **npm and native** install recipes both seed the `/agentbox-setup` skill. Docs cover `--agents` and the Hetzner tier model.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 538d8a35bca2d9a89ab8ad9d15bf78b5f71a3600. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #330, opened 2026-08-29
Cloud B2 (part 1): agentless base + per-agent artifacts for e2b and daytona

madarco:
Stacked on #329. Same three tiers as hetzner — agentless base, one artifact per
agent set on top, runtime install as the fallback — for the two providers whose
derive needs no VPS.

| provider | how a variant derives | measured |
|---|---|---|
| e2b | `Template().fromTemplate(base)` — declarative, no boot | **43s** |
| daytona (container) | recipe appended to the same Dockerfile build; the builder's layer cache serves everything below | **77s** (base: 3m18s) |
| daytona (linux-vm) | boot base snapshot → install → stop → cold-snapshot | see caveat |

## e2b

`build-template.sh` installs no agents and creates no per-agent home/credential
dirs. Prepared state gains a `variants` map (schema 2, lossless migration).

No GC, unlike hetzner: E2B templates are *named*, so rebuilding
`agentbox-claude:latest` moves the tag rather than orphaning an id — and the SDK
exposes no `Template.delete` to reap one with.

## daytona

`prepare --provider daytona --agents claude` previously parsed the flag and
**dropped it** — `prepareDaytona` never read `opts.agents`. Both classes now
derive, by different mechanisms, and a variant takes its class from the BASE
rather than `box.daytonaClass`: a snapshot's class is immutable, so a config key
that disagrees with what was baked would produce an unbootable pairing.

Two daytona-specific constraints shape it:

- **No labels.** `snapshot.create` accepts only name/image/resources/region, so
  unlike hetzner's `agentbox.agents` label the NAME is the only channel saying
  which tier a snapshot is. Variants are `agentbox-<set>-<fp12>`.
- **Never reuse a snapshot name** — Daytona's delete is async and a recreated
  name can report `active` yet fail to boot, so variants carry the same
  `nonce()` the base bake uses.

### The daytona regression from #328 is fixed

After pause/resume a box now has only claude's credential, with no
codex/opencode `auth.json` re-materialising. It was never host re-seeding: the
archive/restore re-applied content from a baked base that carried every agent's
credential paths, and the agentless base removes the source.

## Verified live

**e2b** — agentless base rebuilt (the `agent CLIs` and `Claude Code` build steps
are gone), `agentbox-claude:latest` derived in 43s, box boots it with claude on
the box user's PATH, no codex/opencode, only claude's credential, real
`claude -p` returns OK.

**daytona (container)** — base bakes with the expected missing-VM-image warning;
`--agents claude` derives in 77s, cache-hitting layers 1-62 and running only the
recipe, postInstall and the PATH check; a box boots the variant with no
create-time install and answers a real `claude -p`; `box.imageDaytona` names the
base and `refIsOurBase` still routes to the variant (the trap this was written
for, confirmed live); skip-fast names the claude variant; isolation holds across
pause/resume. All boxes destroyed.

## Caveat

**The daytona `linux-vm` derive is not live-tested.** It needs a published
`ghcr.io/madarco/agentbox/box:sha-<context>`, and the agentless `Dockerfile.box`
has never been published (#328 merged to `feat/custom_agents`, not `main`), so
every local bake downgrades to container:

```
local docker context sha: c79f82835a3ede35
GHCR sha-c79f82835a3ede35 -> 404   (100 tags published, none for this context)
```

Pointing `box.daytonaVmBaseImage` at an existing tag would boot an image that
still carries all three agents, defeating the point. Worth a live pass once an
agentless box image reaches GHCR.

Vercel and DigitalOcean are the second PR.

---

> [!NOTE]
> **Medium Risk**
> Changes cloud box boot paths, prepared-state schemas, and snapshot/template selection for E2B and Daytona; mistakes could boot the wrong artifact or break create after prepare, though migrations are lossless and behavior is covered by new resolution tests.
> 
> **Overview**
> **E2B and Daytona** now follow the same three-tier model as Docker/Hetzner: an **agentless base** (`agentbox prepare --provider …`), optional **per-agent-set bakes** (`--agents claude`), and **on-demand install** when no matching artifact exists.
> 
> **E2B** strips agent CLIs and per-agent credential/home setup from `build-template.sh` (only box runtime like `agent-browser` stays). `prepare` can build derived templates via `Template().fromTemplate(base)` using shared `AGENT_SYNC_SPECS` recipes. Prepared state moves to **schema 2** with a `variants` map (lossless v1 migration). **`provision`** picks the variant template through new `resolveE2bTemplate`, falling back to the agentless base when needed.
> 
> **Daytona** wires up **`--agents`** end-to-end (it was previously ignored). **Container** variants append agent install layers to the same Dockerfile build for layer-cache speed; **linux-vm** variants boot the base snapshot, install, stop, and cold-snapshot (`bakeDaytonaVmVariant`). Prepared state is also **schema 2** with per-variant records; snapshot **names** encode the agent set (no labels API). **`provision`** treats a pinned `box.imageDaytona` that points at the agentless base as “no explicit choice” (`refIsOurBase`) so variant selection still works, and class/env/size come from the snapshot actually booted.
> 
> Docs (`daytona.mdx`, `e2b.mdx`, `agents.md`) describe the split base vs agents and mark the old Daytona pause/resume credential leak as fixed with an agentless base.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 844da976d4058da35f2615d120f27f48e5c938b1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/content/docs/daytona.mdx
+++ b/apps/web/content/docs/daytona.mdx
@@ -46,7 +46,7 @@ The easiest path is the interactive wizard — it takes your API key and bakes t
 agentbox install        # then select daytona
 ```
 
-Paste a **`DAYTONA_API_KEY`** when prompted; the wizard offers to open [the keys page](https://app.daytona.io/dashboard/keys) in your browser first so you can generate one. (There's no browser sign-in — the key is the whole auth. If you paste a **JWT** instead of an API key, you'll also be asked for a `DAYTONA_ORGANIZATION_ID`, which an API key doesn't need: the SDK derives the org from it.) Credentials persist to `~/.agentbox/secrets.env`; project `.env` files are never harvested. `install` also bakes the base snapshot (a one-time `agentbox prepare --provider daytona` under the hood) with the AgentBox runtime — `agentbox-ctl`, the agents, tmux — so every new box boots in seconds instead of eating a ~7-minute cold Dockerfile build.
+Paste a **`DAYTONA_API_KEY`** when prompted; the wizard offers to open [the keys page](https://app.daytona.io/dashboard/keys) in your browser first so you can generate one. (There's no browser sign-in — the key is the whole auth. If you paste a **JWT** instead of an API key, you'll also be asked for a `DAYTONA_ORGANIZATION_ID`, which an API key doesn't need: the SDK derives the org from it.) Credentials persist to `~/.agentbox/secrets.env`; project `.env` files are never harvested. `install` also bakes the base snapshot (a one-time `agentbox prepare --provider daytona` under the hood) with the AgentBox runtime — `agentbox-ctl`, tmux, the browser stack — so every new box boots in seconds instead of eating a ~7-minute cold Dockerfile build.
 
 For CI or scripted setup, run the explicit equivalents:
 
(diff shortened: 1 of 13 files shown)


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #331, opened 2026-08-29
Cloud B2 (part 2): agentless base + per-agent snapshots for DigitalOcean

madarco:
Third and second-to-last piece of the cloud agent-catalog work, after #329 (hetzner) and #330 (e2b + daytona). DigitalOcean now has the same three tiers as every other provider: an **agentless base**, one **derived snapshot per agent set**, and `ensureAgentInstalled` as the runtime fallback. Vercel follows in the next PR.

No CLI-side changes — `--agents`, the hub job payload, and the adoption/pin guards were already provider-generic.

## The provider change

- `install-box.sh` no longer installs claude/codex/opencode or their credential pivots; those live on each agent's `install.postInstall`, so a baked agent and a runtime-added one are byte-identical.
- `prepare --agents <set>` boots the base, runs only those agents' recipes, and re-snapshots.
- Prepared state → schema 3 with a per-variant map and a lossless v2 lift.
- Create picks the variant matching the box's agent set, **falling back to the agentless base**, with the `refIsOurBase` escape so the bake's own `box.imageDigitalocean` pin can't defeat variant selection.
- Per-variant GC. DigitalOcean had **no** base-snapshot GC at all, so every re-bake to date orphaned its predecessor (10 orphans, 51 GB, still on the account).

Three things were genuinely DO-specific rather than a hetzner copy:

| | why |
|---|---|
| derived cloud-init logs in as `vscode` | the base bakes `PermitRootLogin no`, so a root key is accepted by cloud-init then refused by sshd — it surfaces only as an unexplained `waitForSsh` timeout |
| that cloud-init is ASCII-only | DO truncates user-data at the first non-ASCII byte. Hetzner's generator has an em-dash in a comment; copying it verbatim would silently drop the whole document |
| the bake picks a region from `snapshot.regions` | a DO snapshot only boots where it was replicated. Hetzner snapshots are account-wide |

## Three bugs found by live testing

Two are **not DO-specific** and affect already-merged providers.

**1. `~/.claude/skills` was created root-owned** (`sandbox-core`, affects every provider). GNU `install -d -o u -g g a/b/c` applies ownership to the *final component only*, so folding `skills` into the nested `skills/agentbox-setup` call left it owned by root — `postInstall` runs as root. The later static-config stage runs `sudo -u vscode tar -C ~/.claude`, can't create `skills/<name>`, and aborts the bake. Introduced in #329 when the wizard skill moved onto claude's recipe; latent on hetzner and e2b.

**2. Snapshots raced the page cache** (DO **and hetzner**). Both image a *live* machine and neither flushed first. On a derived bake the static-config stage wrote 24 MB and the snapshot was requested **134 ms later**, so every box booted from it was missing the host's Claude skills — while `tar` exited 0 and the bake reported success. Silent data loss with no error anywhere. Both now `sync` and fail loudly rather than snapshot unflushed writes.

**3. Cross-region creates reported the wrong cause.** DO answers a cross-region boot with a generic 422 that we rendered as *"no capacity for s-2vcpu-4gb"*, sending you after a bigger plan when the fix is a region or a re-bake. The create preflight already had the snapshot in hand for the disk check; the region assertion now sits next to it and fails before a billable Droplet exists.

## Verification

Live on DigitalOcean, end to end:

- agentless base baked in **7m34s** (was documented 15–20 min), GC reaped its predecessor
- `--agents claude` derive in **2m42s**, named `agentbox-claude-*`, recorded as its own variant
- `box.imageDigitalocean` stayed pinned to the **base** — a variant bake can't hijack every box
- a real box booted **image 243196710, the variant**, proving `refIsOurBase` + variant selection
- `claude` present, `codex`/`opencode` absent; only `~/.agentbox-creds/claude/` exists
- isolation holds across pause/resume
- skills **1 → 15**, `SKILL.md` **0 → 22,175 bytes**, `~/.agents/skills` **0 → 14** after the sync fix
- no leftover droplets

Unit coverage for the two failure modes that hide from live tests — the base-only fallback (exactly the e2b bug Bugbot caught on #330) and the region mismatch. I confirmed each new test **fails when the fix is reverted** rather than assuming it bites.

`prepare --status` also gained a digitalocean section and per-variant rows, which closes a gap from #330: e2b grew variants but its status never showed them.

## Not in scope

The 10 orphaned DO base snapshots (51 GB) and the 13 on hetzner (~32 GB). Per-variant GC only reaps what it supersedes; a whole-account sweep in `agentbox prune` is separate work.

---

> [!NOTE]
> **Medium Risk**
> Changes cloud image resolution, snapshot baking, and shared agent install recipes used by every provider; mistakes could break creates or produce incomplete snapshots, though behavior mirrors already-shipped hetzner/e2b patterns and adds targeted tests.
> 
> **Overview**
> DigitalOcean now matches the other cloud providers: an **agentless base** snapshot, **per-agent derived snapshots** via `prepare --provider digitalocean --agents <set>`, and create-time selection of the matching variant (with fallback to the base and `ensureAgentInstalled`).
> 
> The DO base `install-box.sh` no longer installs Claude/Codex/OpenCode or credential pivots; those come from `AGENT_SYNC_SPECS` on derive or on-demand install. Prepared state moves to **schema 3** with a `variants` map, per-variant GC, derived bake cloud-init as **`vscode`** (ASCII-only), and region chosen from **`snapshot.regions`**. Create resolves images with **`refNamesPreparedBase`** so `box.imageDigitalocean` pinned to the base does not block variant selection, and preflight surfaces **cross-region** snapshot errors with the correct tier (`base` vs `--agents`).
> 
> **`agentbox prepare` status** adds a DigitalOcean section and lists per-variant rows for E2B and DO (E2B label `tmpl` → `base`).
> 
> Cross-provider fixes: **`~/.claude/skills`** is created with a dedicated `install -d` so GNU install does not leave `skills/` root-owned and break static-config `tar` during bakes (**sandbox-core**, all providers). **Hetzner and DigitalOcean** run **`sync`** before snapshotting so buffered static-config writes are not missing from live snapshots.
> 
> Docs and `docs/agents.md` mark DO as agentless-base + derived variants; new unit tests cover image resolution and region preflight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c6f7aa5a1ca6711e20fcf4e78c8f1be5ae120d6. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

PR #332, opened 2026-08-29
Cloud B2 (final): agentless base + per-agent snapshots for Vercel

madarco:
Final piece of the cloud agent-catalog work, after #329 (hetzner), #330 (e2b + daytona) and #331 (digitalocean). **Every provider now has the same three tiers**: an agentless base, one derived snapshot per agent set, and `ensureAgentInstalled` as the runtime fallback.

No CLI-side changes — `--agents`, the hub job payload and the adoption/pin guards were already provider-generic.

## Vercel was a fork, not a port

| | why |
|---|---|
| Amazon Linux 2023, not Debian | `renderAptInstall` hardcoded `apt-get`, which exits 127 there |
| snapshots carry no name/label/tag | `createSnapshot` takes only `{sessionId, expiration}`, so `vercel-prepared.json` is the *only* record of which snapshot is which |
| `sb.snapshot()` stops before imaging | so vercel is **immune** to the page-cache race fixed for hetzner/DO in #331 — deliberately **no** `sync` added here |

It also needs no `refIsOurBase` escape: the vercel backend never reads `req.image`, so `box.imageVercel` can't defeat variant selection (it's written by the bake but never consulted).

## Two safety fixes this change would otherwise have armed

**1. `destroy` protected exactly one snapshot id.** Everything else relied on `currentSnapshotId === sourceSnapshotId`, but `sourceSnapshotId` is *session*-scoped and can be absent on a resumed session, while `currentSnapshotId` is sandbox-scoped and sticky. Any shared snapshot that wasn't the base could be deleted out from under every box booting it — which **already exposed checkpoints today**, before variants existed. `deleteVercelSnapshot` (the `checkpoint rm` path) had no guard at all. Both now check every locally-known shared id.

**2. Prerequisites were Debian-only.** Codex declares `bubblewrap`; on AL2023 the apt line exits 127 and `ensureAgentInstalled` **throws**, aborting box create — over a package the registry itself calls optional. `renderPackageInstall` now dispatches on the box's actual package manager (apt-get | dnf | microdnf), and prerequisites can be marked `packagesOptional` (codex's is) so a soft dependency logs and continues. New prerequisites default to **required** and must opt into being skippable.

The rename touches 7 call sites across every provider.

## Verification

Live on Vercel, end to end:

- agentless base baked in **2m20s**; GC reaped the superseded v1 base
- `--agents claude` derive in **33s** — fastest of any provider — booting **from the base snapshot**, not the stock runtime
- staging gate held: only claude + `~/.agents` staged, codex/opencode skipped
- a claude box: `claude` present, `codex`/`opencode` absent, only `~/.agentbox-creds/claude/` exists
- **a real `claude -p` turn returned `OK`**
- 15 skills, `SKILL.md` 22,175 bytes `vscode:vscode`, `~/.agents/skills` 14 — vercel never staged `~/.agents` before this
- isolation holds across pause/resume
- **the destroy guard, live**: destroyed a box booted from the claude variant and a box booted from the base; both shared snapshots still skip-fast (which probes the snapshot remotely), so neither was deleted

Two paths my earlier live passes would have skipped, and which found bugs on the other providers:

- **base-only fallback** — the codex box has no variant, so it booted the base and installed at create. Exactly the shape of the high-severity e2b bug on #330.
- **the AL2023 prerequisite path** — `agentbox codex --provider vercel` on the agentless base. This is the create that would have aborted. It succeeded, and answered a question the repo had no evidence for: **AL2023 does carry bubblewrap** (`/usr/bin/bwrap` present), so the dnf branch installed it and the optional fallback never had to fire.

Unit coverage for what live runs hide: the destroy guard against a variant id and a checkpoint id, the base-only create fallback, the schema-1→2 migration, the derived builder passing no `runtime` and no `keepLastSnapshots`, and the optional-vs-required prerequisite branch. I confirmed each **fails when the fix is reverted** — three guard tests and the optional-prereq test.

`prepare --status` also gains **vercel and hetzner** sections, so every provider with variants is now visible.

---

> [!NOTE]
> **High Risk**
> Changes Vercel provisioning, snapshot GC, and destroy/checkpoint deletion guards—mistakes could delete shared snapshots other boxes depend on; install path changes affect all providers but are covered by tests.
> 
> **Overview**
> Brings **Vercel** in line with the other clouds: an **agentless** base snapshot, **`prepare --agents`** derived snapshots keyed in `vercel-prepared.json` (schema 2 + `variants`), and **create** booting the matching variant or falling back to the base so `ensureAgentInstalled` can run.
> 
> **Install plumbing** renames agent **`apt`** prerequisites to **`packages`**, replaces **`renderAptInstall`** with **`renderPackageInstall`** (apt-get / dnf / microdnf) across docker and all cloud prepare paths, and treats Codex’s **`bubblewrap`** as **`packagesOptional`** so AL2023 box creates don’t fail when the package is missing.
> 
> **Vercel-specific safety**: **`resolveVercelSnapshot`**, **`sharedSnapshotIds`**, and stricter **`destroy`** / **`deleteVercelSnapshot`** so prepared base/variant ids and checkpoint manifests aren’t deleted when a box is torn down. **`provision.sh`** drops baked-in agent CLIs and per-agent credential dirs (those come from catalog **`postInstall`** / derived bake). Derived builders boot from the base snapshot with **no** `keepLastSnapshots` that would evict the shared base.
> 
> **CLI/docs**: `agentbox prepare` status adds **hetzner** and **vercel** sections; docs describe agentless Vercel prepare and per-agent bakes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2ee3f8fbcdfa72ff4620b3e706919908d9eed48. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #333, opened 2026-08-29
Checkpoints record their agent set; honest box.image<Provider> docs

madarco:
Two loose ends from the cloud agent-catalog work (#329–#332). Both are metadata-honesty problems, so they land together.

## 1. Checkpoints record their agent set

A checkpoint captured from a claude box restored silently into an `agentbox codex` box. The sharp case is `--set-default`, which writes a **per-provider** key: a default captured from one agent's box applied to *every* box on that provider, **with no user signal at all**. Every backend returns early on `req.snapshot` before it ever looks at `req.agents`, and nothing filtered a checkpoint by agent set.

Both manifests now carry an optional `agents`, and all four create paths warn when it differs. Per the chosen behaviour it is **advisory** — the checkpoint still boots, and the missing agent is installed at create.

Two decisions worth review attention:

**No schema bump**, deliberately the opposite call from the prepared-state bumps in #329–#332. There, an older CLI that ignored `variants` would boot the *wrong image*, so making the file unreadable was the safer failure. Here the field is advisory: the readers hard-gate on the schema number, so bumping would make every new checkpoint **vanish** from an older CLI's `agentbox checkpoints`, while an older CLI that ignores the field behaves exactly as today. Losing a warning is much cheaper than losing sight of a checkpoint. There's a test pinning that an unknown future field doesn't break reads, since that's the assumption this rests on.

**The warning lives at ref resolution, not in the setup wizard.** I initially put it in the wizard — then found `codex` and `opencode` never run the wizard; they read `box.defaultCheckpoint<Provider>` straight into `checkpointRef`. A wizard-only warning would have missed the exact case this exists for.

`agentMismatch` rides alongside `CheckpointStatus` rather than folding into `stale`, because a checkpoint can be perfectly fresh and still be the wrong agent — overloading `stale` would make an up-to-date checkpoint look outdated and invite the wizard to offer discarding it.

**Absent always means UNKNOWN, never "no agents."** Every pre-existing checkpoint lacks the field; treating that as a mismatch would warn on every restore.

## 2. `box.imageVercel` / `box.imageE2b` are written but never read

Neither backend consults `req.image` — both resolve from their own prepared state, and their snapshots/templates are opaque ids with no other addressing. Behaviour is unchanged; the descriptions now say so, and `prepare --status` no longer prints `(pinned in project)` for e2b, which claimed a control the user doesn't have. (Vercel already suppressed it.)

## Bugs found while doing this

- **Docker never persisted `BoxRecord.agents`.** The field is documented as "the agents this box was created FOR" and the cloud path sets it; docker carried only `lastAgent`, which answers a different question (whichever agent most recently *ran*). Caught because the new manifest field came out empty on docker. `opts.agents` was already in scope — it picks the derived image variant a few lines above.
- **`sandbox-cloud`'s checkpoint suite was writing into the real `~/.agentbox`.** `CLOUD_CHECKPOINTS_ROOT` is fixed at module load, so the `beforeAll` HOME override never moved it. Every run left manifest dirs in the developer's home. Tests now locate manifests via the module's own exported root and clean up after themselves.

## Verification

Live on docker, the full scenario:

```
checkpoint create ckptclaude2 --name warm2 --set-default   → manifest: agents: ['claude']
codex -y -d -n ckptcodex                                    → ▲ "checkpoint was captured from a
                                                               claude box but this box is for codex…"
                                                            → still boots from the checkpoint
```

Then the regression that matters most: hand-edited `agents` out of the manifest, re-ran the codex create, and confirmed **zero** warnings — a legacy checkpoint stays silent.

Unit coverage for absent-is-not-a-mismatch, order-insensitivity, manifest round-trip on both stores, legacy manifests still loading, and unknown-future-field tolerance. I verified the absent case **fails when the guard is removed** rather than assuming it bites.

---

> [!NOTE]
> **Low Risk**
> Changes are mostly advisory warnings and optional manifest metadata; boot behavior is unchanged, with explicit guards so legacy checkpoints and failed reads stay silent.
> 
> **Overview**
> **Checkpoints now record which agent set they were captured from** and the CLI **warns (advisory, still boots)** when a resolved checkpoint ref — especially a per-provider default — was taken from a different agent than the box being created.
> 
> Manifests on **Docker and cloud** gain an optional `agents` field **without a schema bump** so older CLIs keep listing checkpoints; absent/empty means unknown (no warning). **`evaluateCheckpoint`** can attach **`agentMismatch`** alongside fresh/stale; **`warnCheckpointAgentMismatch`** reads manifests only (no cloud liveness probe) and is invoked at **ref resolution** in **`create`**, **`claude`**, **`codex`**, and **`opencode`** (including **before** the `-i` queue early-return).
> 
> **Docker create** now persists **`BoxRecord.agents`** so checkpoint capture can record the intended agent set (cloud already did).
> 
> **Docs / prepare status:** **`box.imageVercel`** and **`box.imageE2b`** are documented as written by prepare but **not read** (backends use prepared state); **`prepare --status`** drops the misleading e2b “pinned in project” line.
> 
> Tests cover mismatch rules, warn ordering, manifest round-trip/legacy tolerance, and cloud checkpoint test cleanup under real home paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bec337b26642d4cc04f858c7b5e3177645bd51db. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

PR #338, opened 2026-08-30
feat(agents): derive the download paths and item lists from the registry

madarco:
Phase 2 of the agent data plane (backlog item 1 from #336). **Lands narrower than the backlog entry describes** — see "What's still open".

## The problem

The box→host direction restated what host→box already knew: box-dir literals in `agent-pull.ts` alongside the same values in `registry.ts`, plus the pull item lists as module constants — with nothing to catch a divergence.

## What's fixed

- `agentBoxDir(agent)` derives the root from `staticPaths[0].boxDir`, so the two directions can no longer disagree.
- Each spec carries an `AgentPullSpec` (`items` / `categories` / `jsonMerges`) that `CLAUDE_PULL_DIR_CATEGORIES`, `CODEX_PULL_ITEMS` and the two opencode lists derive from.
- A drift test asserts the derivation, and that **every agent declaring `staticPaths` also declares how to pull them** — the gap that let `download` be forgotten silently when adding an agent (nothing fails without it; the subcommand is simply absent).

I verified the derived values are **byte-identical** to the literals they replace — the real regression check for a refactor with no intended behaviour change.

## Why `AgentPullSpec` is not just `staticPaths`

Two things make reuse wrong, both pinned by tests:

- That field's `exclude`/`include` are **push-direction hygiene** — claude drops `projects`/`sessions` on the way *in*, while the pull's real filters are `node_modules` and the `agentbox-` skill prefix. Reusing them would be wrong in both directions.
- Opencode's third root (`~/.local/state/opencode`, `update: true`) must **never** be pulled: newest-wins two-way state is the opposite of pull's additive never-overwrite rule.

## What's still open (left in the backlog, not dropped)

The three `download-<agent>.ts` commands and the three `pull*ViaTransport` functions are **not** collapsed. Codex and opencode are the same function differing only in group/root/items, but claude is a genuinely different *shape* — category children plus a JSON registry merge with a container→host path rewrite — so collapsing needs both strategies actually implemented, not just declared.

I stopped deliberately: `download` writes to the user's real `~/.claude`, and turning "additive" into "overwrite" destroys host config. Landing the verified half beat rushing the whole.

## Live verification (docker)

A planted claude skill is enumerated, and on a real opencode box `config/agents` + `config/themes` are both detected — the latter exercises the `relocToSubpath` config group end to end.

Worth recording: my first probe run put codex/opencode items in a *claude* box and got "no new" for both, which looked like a bug. It wasn't — per-agent isolation means only `agentbox-claude-config` is mounted, so those writes never reached the volumes `download` reads. Confirmed with `docker inspect`, then re-tested against a real opencode box.

---

> [!NOTE]
> **Low Risk**
> Intended behavior-preserving refactor on additive download paths, with drift tests and explicit exclusion of newest-wins OpenCode state from pull.
> 
> **Overview**
> **Phase 2 of the agent download backlog:** box→host sync no longer duplicates path and item knowledge that already lives in `AGENT_SYNC_SPECS`.
> 
> Each agent row now has an optional **`AgentPullSpec`** (`items`, `categories`, `jsonMerges`) describing how `agentbox download` inventories the box. **`agentBoxDir`**, **`pullItems`**, and **`pullCategories`** in `agent-pull.ts` read that data (box roots from `staticPaths[0].boxDir`; lists from `pull`), so exports like `CLAUDE_PULL_DIR_CATEGORIES` and `CODEX_PULL_ITEMS` are derived instead of hardcoded literals.
> 
> A new **`agent-pull-drift.test.ts`** asserts those derivations stay aligned with the registry, that every agent with `staticPaths` also defines a non-empty `pull` spec (so new agents can’t silently ship without download), and that OpenCode’s `update: true` state root is excluded from pull. **`docs/agents.md`** marks backlog item 1 as partly done and notes what’s still open (separate download CLI commands, three `pull*ViaTransport` implementations, docker inventory duplication).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d818937524eee6f720ba1537bccaea4966f43865. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -332,22 +332,31 @@ baked agent and a runtime-added one are byte-identical. Everything below is a
 place where agent knowledge still lives as **code** instead of data. None of it
 is broken; all of it is work a fourth agent would pay for.
 
-### 1. `agentbox download` duplicates the registry's paths, in reverse
+### 1. `agentbox download` duplicates the registry's paths, in reverse — **partly done**
 
-`packages/sandbox-core/src/sync/agent-pull.ts` hardcodes `CLAUDE_BOX_CONFIG_DIR`
-/ `CODEX_BOX_CONFIG_DIR` / `OPENCODE_BOX_DATA_DIR` and exposes per-agent
-functions (`pullClaudeExtrasViaTransport` and friends). It never reads
-`staticPaths` from `AGENT_SYNC_SPECS` — so the host→box direction is
-registry-driven and the box→host direction is not, with two copies of the same
-path knowledge free to drift.
+Fixed: the box roots and the pull item lists now derive from the registry.
+`agentBoxDir(agent)` reads `staticPaths[0].boxDir`, and each spec carries an
+`AgentPullSpec` (`items` / `categories` / `jsonMerges`) that
+`CLAUDE_PULL_DIR_CATEGORIES`, `CODEX_PULL_ITEMS` and the two opencode lists are
+derived from. A drift test asserts the derivation and that every agent declaring
+`staticPaths` also declares how to pull them — the gap that let `download` be
+forgotten silently.
 
-`apps/cli/src/commands/download-{claude,codex,opencode}.ts` are ~120 lines each
-over shared `_agent-pull.ts` / `_agent-propagate.ts` helpers.
+`AgentPullSpec` is deliberately NOT `staticPaths`: that field's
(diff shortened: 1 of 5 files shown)


madarco:
bugbot run

PR #339, opened 2026-08-30
feat(agents): ctl pulls its watch list from the host over RPC

madarco:
Phase 3 of the agent data plane, and the first half of backlog item 2 from #336. **Stacked on #338** (base is `feat/agent-data-plane-pull`).

## The problem

`agentbox-ctl` is baked into the box image, so its watch list is frozen at bake time. A box built from a snapshot taken before an agent existed never watches that agent's credential — and a plugin-supplied agent, which is `ondemand`-only and can *never* be baked, would be invisible to ctl forever.

## RPC, not a pushed file

ctl now reconciles against an `agents.list` RPC at daemon start, modelled on `tool.list`: read-only, unprompted, and dispatched **before worktree resolution** since the agent registry is machine-global rather than project-scoped.

I originally planned to push `/etc/agentbox/agents.json` at create. Marco pushed back, correctly: a file has to be *written by someone*, and that someone is the **provider** — cloud bootstrap, docker box-env, and a hand-written equivalent in every community provider. The file's shape then becomes a contract each of them implements, so any future field means updating all of them, third-party ones included. Over RPC the shape stays host-side and providers stay entirely uninvolved.

## Both dispatch sites

Handled in `server.ts` (docker) **and** `host-actions.ts` (cloud). A cloud box's RPCs are parked on the `HostActionQueue` and drained by a separate executor, so a method wired only in `server.ts` works on docker and silently fails on every cloud provider.

## Safety posture

- **Ships a projection, not the raw table.** `credential.hostBackup` is an absolute *host* path baked at module load — meaningless inside a box and a leak of the host's home layout. Pinned by a test.
- **Every failure keeps the baked list.** Unreachable relay, older host with no `agents.list`, malformed JSON, and an **empty `agents` array** are all refused. Taking an empty list literally would stop the credential watcher dead and kill login fan-out across the fleet with no error anywhere — so "watch nothing" is treated as malformed, never as authoritative.
- **Unknown fields are ignored**, so a newer host never breaks an older box — the same posture as the in-box `agentbox.yaml` parser.
- **A `fanout` watch with no validatable shape is dropped**, not posted unvalidated.

## The hooks a custom agent declares

- `watch` — extra in-box files, with **`backup` the default and `fanout` opt-in**. Fanout re-distributes to *every other box*, which is correct only for a rotating secret; doing it to session transcripts would be a privacy leak.
- `roots` on `AgentPullSpec` — pull-only locations. Session logs are genuinely pull-only: you'd never *push* transcripts into a box, and every agent's `staticPaths.exclude` already drops them (claude `projects`/`sessions`/`history.jsonl`, codex `sessions`/`archived_sessions`/`log`, opencode `storage`/`log`/`snapshot`). With groups resolving only through `staticPaths` there was nowhere to declare them — the gap Marco identified in #338.

## Verification

Live on docker, the full loop:

```
agentbox-hub: rpc box=bd40b86a5 method=agents.list
agentbox-hub: credentials-updated box=bd40b86a5 accepted=false (unchanged)
```

ctl pulled the list over RPC and the watcher then ran off it, with the newest-wins gate correctly rejecting unchanged content. I also queried `agents.list` directly from inside the box and got the full payload.

Unit coverage for the refusal cases, and I confirmed the two safety-critical ones **fail when the guard is removed** — the empty-list refusal and the host-path leak.

## Not in this PR

- **Refreshing an already-running box.** `ToolLinksWatcher` solves the equivalent with `agentbox-ctl tool relink` over `Provider.exec`, but that works because tool links are on-disk state a separate process can rewrite; this list lives in the daemon's memory, so the same push needs a ctl socket op. A running box picks up changes on the next ctl restart — which is still the win: adding an agent goes from "re-bake every base" to "restart ctl".
- Consuming the `watch`/`roots` hooks end-to-end (the `cp.toHost` backup route, `box.agentFileSync`) — declared and shipped here, wired next.
- `boxRunEnv` is still declared-and-unconsumed.

---

> [!NOTE]
> **Medium Risk**
> Changes credential watcher startup and fleet-wide login fan-out behavior; failures are designed to fall back to the baked list, but a bad host payload that parses incorrectly could still alter which paths are watched until restart.
> 
> **Overview**
> **agentbox-ctl** no longer relies solely on the credential paths compiled into the box image. At daemon startup it calls a new read-only **`agents.list`** RPC, uses the returned **fanout** credential watches for **`CredentialsWatcher`**, and on any failure (unreachable relay, bad JSON, empty list, older host) keeps the existing baked list so credential fan-out never goes silent.
> 
> The host exposes **`buildAgentDescriptors()`** — a projection of **`AGENT_SYNC_SPECS`** without host-only paths — from **`server.ts`** (docker) and **`host-actions.ts`** (cloud queue). In-box **`parseAgentDescriptors`** tolerates unknown fields and only admits validated **fanout** shapes for the watcher.
> 
> Agent sync types gain optional **`watch`** (extra in-box files, default **backup**) and **`pull.roots`** for pull-only locations; those hooks are declared and shipped on the wire but not fully consumed in ctl yet. Running boxes pick up registry changes on the next ctl restart, not live.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c21b96132a59fdbc5514c401a5ed51b6e0b765bf. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ctl/src/agent-registry.ts
+++ b/packages/ctl/src/agent-registry.ts
@@ -1,0 +1,102 @@
+import { postRpcAwait } from './relay-rpc.js';
+import { WATCHED_CREDENTIALS, type WatchedCredential } from './credentials-watcher.js';
+
+/**
+ * The agent watch list, pulled from the host over `agents.list`.
+ *
+ * Why this exists: `agentbox-ctl` is BAKED INTO THE BOX IMAGE, so its compiled-in
+ * list is frozen at bake time. A box built from a snapshot taken before an agent
+ * existed never watches that agent's files. A plugin-supplied agent is
+ * `ondemand`-only and can never be baked at all, so without this it would be
+ * permanently invisible to ctl.
+ *
+ * Pulled rather than pushed as a file: a file has to be written by the PROVIDER
+ * (cloud bootstrap, docker box-env, and a hand-written equivalent in every
+ * community provider), which makes its shape a contract each of them implements.
+ * Over RPC the shape stays host-side and providers stay uninvolved.
+ *
+ * Reconciled ONCE at daemon start, never polled — on a control box a poll would
+ * be a WAN round trip per box per minute (the reason `ToolLinksWatcher` stopped
+ * polling too). That covers a new box, a restarted box and a resumed box.
+ *
+ * NOT yet covered: a box already running when the host's agent list changes; it
+ * picks the change up on the next ctl restart. `ToolLinksWatcher` solves the
+ * equivalent with `agentbox-ctl tool relink` over `Provider.exec`, but that works
+ * because tool links are on-disk state a separate process can rewrite — this list
+ * lives in the daemon's memory, so the same push needs a ctl socket op. Left for
+ * a follow-up; the startup reconcile is what turns adding an agent from
+ * "re-bake every base" into "restart ctl".
+ *
+ * FAILURE IS SILENT AND KEEPS THE BAKED LIST. An unreachable relay, an older
+ * host with no `agents.list`, or a malformed payload must all leave the box
+ * watching what it already watched — never nothing. Losing the credential
+ * watcher would silently break login fan-out for the whole fleet.
+ */
+
+interface WireWatch {
+  path?: unknown;
+  sync?: unknown;
+  shape?: unknown;
+  hostDest?: unknown;
+}
+interface WireAgent {
+  id?: unknown;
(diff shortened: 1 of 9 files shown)


madarco:
bugbot run

madarco:
bugbot run

PR #340, opened 2026-08-30
feat(agents): ctl pulls its watch list from the host over RPC

madarco:
Phase 3 of the agent data plane. **Replaces #339**, which GitHub auto-closed when I merged #338 with `--delete-branch` — that deleted #339's base branch. Same commits, rebased onto `feat/custom_agents`; the review history is on #339.

Backlog item 2 from #336.

## The problem

`agentbox-ctl` is baked into the box image, so its watch list is frozen at bake time. A box built from a snapshot taken before an agent existed never watches that agent's credential — and a plugin-supplied agent, which is `ondemand`-only and can *never* be baked, would be invisible to ctl forever.

## RPC, not a pushed file

ctl reconciles against an `agents.list` RPC at daemon start, modelled on `tool.list`: read-only, unprompted, dispatched **before worktree resolution** since the registry is machine-global.

I originally planned to push `/etc/agentbox/agents.json`. Marco pushed back, correctly: a file has to be *written by someone*, and that someone is the **provider** — cloud bootstrap, docker box-env, and a hand-written equivalent in every community provider. Its shape becomes a contract each of them implements, so any future field means updating all of them. Over RPC the shape stays host-side and providers stay uninvolved.

## Both dispatch sites

`server.ts` (docker) **and** `host-actions.ts` (cloud). Cloud RPCs are parked on the `HostActionQueue` and drained by a separate executor, so a method wired only in `server.ts` works on docker and silently fails on every cloud provider.

## Safety posture

- **A projection, not the raw table.** `credential.hostBackup` is an absolute *host* path baked at module load — meaningless in a box and a leak of the host's home layout. Pinned by a test.
- **Every failure keeps the baked list.** Unreachable relay, older host, malformed JSON, and an **empty `agents` array** are all refused. Taking an empty list literally would stop the credential watcher dead and kill login fan-out across the fleet with no error anywhere.
- **Unknown fields ignored**, so a newer host never breaks an older box.
- **A `fanout` watch with no validatable shape is dropped**, not posted unvalidated.

## The hooks a custom agent declares

- `watch` — extra in-box files, **`backup` default, `fanout` opt-in**. Fanout re-distributes to every other box; correct only for a rotating secret, never for transcripts.
- `roots` on `AgentPullSpec` — pull-only locations. Session logs are genuinely pull-only: every agent's `staticPaths.exclude` already drops them, so there was nowhere to declare them.

## The bug Bugbot caught on #339 (fixed here)

**High severity, and the feature was inert.** I had placed `fetchWatchList()` *before* the relay forwarder binds :8788, so the RPC was a guaranteed `ECONNREFUSED` on a fresh box. Because the fallback is silent **by design**, that failure was invisible — the box quietly kept its baked list forever.

The daemon already had this exact rule for `toolLinks.start()`, with the reason in a comment three lines from where I put the fetch. So this adds a **source-level ordering test** rather than another comment, and I confirmed it fails when the fetch is moved back.

**My earlier "live verified" claim on #339 was also wrong** — the `agents.list` line I saw in the hub log came from a manual `curl` I ran inside the box, not from ctl. Re-verified properly: snapshot the hub log, create a box, run no curl, and confirm one `agents.list` from ctl's own startup reconcile. That's what's in the log now.

## Not in this PR

- **Refreshing an already-running box.** `ToolLinksWatcher` does this with `tool relink` over `Provider.exec`, but that works because tool links are on-disk state; this list lives in daemon memory, so it needs a ctl socket op. A running box picks changes up on the next ctl restart — still the win: "re-bake every base" becomes "restart ctl".
- Consuming the `watch`/`roots` hooks end-to-end (`cp.toHost` backup route, `box.agentFileSync`).
- `boxRunEnv` and `caps` remain declared-and-unconsumed.

---

> [!NOTE]
> **Medium Risk**
> Changes credential fan-out startup and new relay paths; regressions could leave post-bake or plugin agents unwatched, though baked-list fallbacks, timeouts, and ordering tests reduce that risk.
> 
> **Overview**
> **In-box ctl** no longer depends only on the image-baked credential paths: at daemon start it calls **`agents.list`** over the in-box relay, parses fanout watches, and upgrades **`CredentialsWatcher`** via **`setFiles`**. The watcher still **starts immediately on the baked list**; the fetch is **`void`** (not awaited), **30s-bounded**, and runs **only after** the relay forwarder listens—fixing the silent ECONNREFUSED case where reconcile never ran. Any timeout, RPC failure, malformed JSON, or empty agent list **keeps the baked list** (never “watch nothing”).
> 
> The **host** adds a read-only **`agents.list`** handler in both **docker relay** (`server.ts`) and **cloud action executor** (`host-actions.ts`), returning **`buildAgentDescriptors()`**—a wire-safe projection of **`AGENT_SYNC_SPECS`** (no host backup paths). Agent sync types gain declarative **`watch`** and pull-only **`roots`** hooks for custom agents; end-to-end backup/fanout beyond credentials is left for follow-ups.
> 
> Tests cover payload parsing, fetch timeout/fallback, **`setFiles`** behavior, and **source-level daemon startup ordering** (watcher before fetch, no `await fetchWatchList`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6776d020b0e02b7c98793c65b79cbe0919e309fb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/ctl/src/agent-registry.ts
+++ b/packages/ctl/src/agent-registry.ts
@@ -1,0 +1,102 @@
+import { postRpcAwait } from './relay-rpc.js';
+import { WATCHED_CREDENTIALS, type WatchedCredential } from './credentials-watcher.js';
+
+/**
+ * The agent watch list, pulled from the host over `agents.list`.
+ *
+ * Why this exists: `agentbox-ctl` is BAKED INTO THE BOX IMAGE, so its compiled-in
+ * list is frozen at bake time. A box built from a snapshot taken before an agent
+ * existed never watches that agent's files. A plugin-supplied agent is
+ * `ondemand`-only and can never be baked at all, so without this it would be
+ * permanently invisible to ctl.
+ *
+ * Pulled rather than pushed as a file: a file has to be written by the PROVIDER
+ * (cloud bootstrap, docker box-env, and a hand-written equivalent in every
+ * community provider), which makes its shape a contract each of them implements.
+ * Over RPC the shape stays host-side and providers stay uninvolved.
+ *
+ * Reconciled ONCE at daemon start, never polled — on a control box a poll would
+ * be a WAN round trip per box per minute (the reason `ToolLinksWatcher` stopped
+ * polling too). That covers a new box, a restarted box and a resumed box.
+ *
+ * NOT yet covered: a box already running when the host's agent list changes; it
+ * picks the change up on the next ctl restart. `ToolLinksWatcher` solves the
+ * equivalent with `agentbox-ctl tool relink` over `Provider.exec`, but that works
+ * because tool links are on-disk state a separate process can rewrite — this list
+ * lives in the daemon's memory, so the same push needs a ctl socket op. Left for
+ * a follow-up; the startup reconcile is what turns adding an agent from
+ * "re-bake every base" into "restart ctl".
+ *
+ * FAILURE IS SILENT AND KEEPS THE BAKED LIST. An unreachable relay, an older
+ * host with no `agents.list`, or a malformed payload must all leave the box
+ * watching what it already watched — never nothing. Losing the credential
+ * watcher would silently break login fan-out for the whole fleet.
+ */
+
+interface WireWatch {
+  path?: unknown;
+  sync?: unknown;
+  shape?: unknown;
+  hostDest?: unknown;
+}
+interface WireAgent {
+  id?: unknown;
+  watch?: unknown;
+}
+
+/**
+ * Narrow the wire payload, dropping anything malformed rather than throwing.
+ *
+ * Unknown fields are IGNORED, not rejected: a newer host must never break an
+ * older box. Same posture as the in-box `agentbox.yaml` parser, which warns on
+ * unknown keys instead of failing.
+ */
+export function parseAgentDescriptors(stdout: string): WatchedCredential[] | null {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(stdout);
+  } catch {
+    return null;
+  }
+  if (parsed === null || typeof parsed !== 'object') return null;
+  const agents = (parsed as { agents?: unknown }).agents;
+  if (!Array.isArray(agents)) return null;
+
+  const out: WatchedCredential[] = [];
+  for (const raw of agents as WireAgent[]) {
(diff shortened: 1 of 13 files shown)


madarco:
Fixed the HIGH — and it was worse than "hang startup", so worth recording the mechanism.

**Root cause.** `relay/src/server.ts:942` parks *every* method except `browser.open` on `HostActionQueue` in box mode, and that branch sits **above** the `agents.list` handler at :954 — so that handler only ever served the docker path. `HostActionQueue.enqueue` is deliberately timeout-free and expiry runs only inside `drain()`, which only runs when the host's `CloudBoxPoller` polls. `relay-rpc`'s own 10-minute bound belongs to the 202/poll shape, not this one. Host off → nothing settles, ever.

Two consequences, not one:
1. `CredentialsWatcher` was never constructed — credential fan-out dead. A regression against `main`, where it started unconditionally.
2. `SIGTERM`/`SIGINT` never registered, so a stop in that window skipped the whole graceful path.

**Fix:** start the watcher on the baked list immediately and upgrade via `setFiles()` when the host answers — the shape `toolLinks.start()` (`void this.tick()`) already uses. Plus a 30s bound on the fetch, since an unbounded pending request holds a socket and a queue slot for the daemon's life.

Not fixable relay-side: `agents.list` answers from the *host's* registry, which is the whole point — the box's baked list is the stale thing.

**Verified live**, not just by unit test — real `agentbox-ctl daemon` against a stand-in relay that accepts `POST /rpc` and never answers (what a box-mode relay does with a dead poller):
- fixed: `agents.list timed out; keeping the baked watch list` at 30s, and a SIGTERM sent at t+5s (mid-hang) logs `SIGTERM — shutting down` and exits cleanly.
- pre-fix, same scenario: **no shutdown line at all** — node's default action killed it, no flush/stopAll/relay-close.

Tests: mutation-checked both new guards (removing the race hangs the timeout test; making `setFiles` a no-op fails the swap test). `daemon-startup-order.test.ts`'s second case asserted the *old* design ("the watcher reads `files` once, at construction") — inverted, since `setFiles()` retires that premise.

bugbot run

madarco:
bugbot run

PR #342, opened 2026-08-30
refactor(agents): one open AgentId, and the per-agent module table

madarco:
Plan items **1+2** of "agents as in-monorepo plugins": converge the agent-name types, open `AgentId`, and add the per-agent module table.

## What the problem actually was

The plan said "`AgentId` is a closed union". Exploration said something worse: **`AgentId` was barely used.** Eight types spelled the same thing —

| type | home |
| --- | --- |
| `AgentKind` | `packages/core/src/types.ts` |
| `SyncAgentKind` / `QueueAgentKind` | `packages/core/src/sync/agent-kind.ts` |
| `AgentId` | `packages/sandbox-core/src/sync/agents/types.ts` |
| `AgentName` | `apps/cli/src/lib/agent-login-specs.ts` |
| `TeleportAgent` | `apps/cli/src/session-teleport/types.ts` |
| `CloudAgentKind` | `packages/sandbox-cloud/src/sync/agent-credentials.ts` |
| `ForkAgent` / `CmuxAgentMode` / `HerdrAgentMode` | `apps/cli/src/**` |

— plus ~35 more sites re-declaring `'claude' | 'codex' | 'opencode'` inline. Opening any one of them changes nothing for the other seven.

## 1. One open type

`AgentId` / `AgentMode` / `QueueAgentKind` now live once, in `@agentbox/core` (the only package with no internal deps, so `sandbox-*`, `relay`, `ctl`, the hub and the CLI can all reach it). Every site above points at them.

`AgentId` is an **open `string`**: an agent is a registry row, so which ids are valid is a runtime fact. `isAgentKind` (`@agentbox/core`) answers for the built-ins — a dependency-free leaf has no registry to ask — and `isRuntimeAgent` (`@agentbox/sandbox-core`) answers from the registry. Same split as `isProviderKind` vs `isRuntimeProvider`.

`ctl`'s `WatchedCredential.agent` opens too, which drops a cast that was a lie: `parseAgentDescriptors` did `id as WatchedCredential['agent']` on ids arriving over `agents.list` — exactly the post-bake agents #340 exists to deliver.

## 2. The module table

`apps/cli/src/agents/<id>/` is now one folder per agent, holding the code the JSON-serializable registry row cannot carry: the guided-login detector and the session-teleport resolver. `prepareTeleport` reads the table instead of switching on the agent name; `session-teleport/{claude,codex}.ts` moved there.

Literal `import()` specifiers, per `provider/loaders.ts` — and verified in the **packed** CLI, not just the dev tree, since that failure mode only appears there.

Deliberately **not** done: `AGENT_SYNC_SPECS` does not derive from this table. The table is lazy and the specs are read synchronously in ~40 places; deriving would make all of them async for no gain (a spec is small pure data, not a heavy SDK). The registry stays the sync source of truth; the table is the behavior tier above it.

## Paying for the open type

The compiler no longer enumerates the sites a new agent misses. Three guards replace it, **each verified to fail when the path it covers is broken**:

- `no-inline-agent-union.test.ts` — no file re-declares the agent set. Tuned to the *agent set*, not to any union containing two agent names: `OpenInApp` legitimately lists `Claude.app` and `Codex.app` beside herdr, cmux and finder, and a fourth agent does not belong there. Flagging it would train people to ignore the test.
- `agent-module-table.test.ts` — keys equal `agentIds()`; each module carries its registry row **by reference**; a `teleport` resolver exists exactly when `caps.teleport` is `'full'`.
- `agent-command-coverage.test.ts` — same for `fork`'s delegate map.

Deleting any of them silently re-closes the type; the doc comment in `agent-kind.ts` says so.

## Found along the way

The sweep surfaced five more places where a fact was stated a second time as a literal, all now derived:

- `agent-sessions`' `ResumableAgent` and `_cloud-attach`'s resume gate → `caps.resume`
- `attach`'s `AGENT_PRIORITY` map → registry order
- `fork`'s `--agent` choices → `agentIds()`
- the custody store's valid `agents/` segments → `agentIds()` (it would have rejected any agent added after it was written)
- `sandbox-cloud`'s forwarded env keys → `spec.forwardedEnvKeys`, which also drops a `sandbox-cloud → sandbox-docker` import

## Verification

- Full CI locally: `check:plugin-skill`, `lint`, `build`, hub `build:standalone`, `check:cloud-backends`, `typecheck`, `test` — all green. 1258 CLI tests pass; no test was dropped (verified per-file against the base branch).
- Each new guard mutation-tested: re-added an inline union, dropped a table arm, dropped a fork arm, gave opencode a teleport resolver it does not declare. All four failed as intended.
- **Packed install** (`npm pack` → isolated `npm i`): `fork --agent bogus` lists the registry agents, `fork --agent opencode --session` refuses on `caps.resume`, `opencode -c` refuses with `caps.teleportStubReason`, and `claude --resume <uuid>` reaches `resolveClaudeTeleport` — proving the lazy agent module resolves in the published bundle.
- Live docker: created a plain box and a `claude -y -d` box; `list` showed `claude:idle`, `download claude --dry-run` and `open --targets` both correct; destroyed both.

## Notes

- Provider SDK **2.10.0 → 2.11.0**: `AgentSyncSpec.id` widened from a union to `string` on the published surface. No `SDK_API_VERSION` bump — the runtime contract is unchanged, so plugins built against 2.10.0 keep working. **Publish is manual.**
- No public-docs change: no CLI command, flag, config key, default or provider behavior changed. `docs/agents.md` is updated (the "adding a new agent" checklist loses the open-the-unions step and gains the `agents/<id>/` folder).

---

> [!NOTE]
> **Medium Risk**
> Touches session restore, attach/fork dispatch, teleport, and credential custody validation across CLI, hub, and relay; mistakes in open-type wiring could fail silently without the new coverage tests.
> 
> **Overview**
> Converges **eight duplicate agent-name types** and dozens of inline `'claude' | 'codex' | 'opencode'` unions onto **`AgentId` / `AgentMode` / `QueueAgentKind`** in `@agentbox/core` as open strings, with registry-backed validation via `isRuntimeAgent` and `agentIds()` where the compiler no longer proves exhaustiveness.
> 
> Adds **`apps/cli/src/agents/<id>/`**: a lazy, literal-`import()` **module table** for per-agent **login detectors** and **teleport resolvers** (moved out of `agent-login-specs.ts` and `session-teleport/`). **`prepareTeleport`** and guided login now load behavior through **`loadAgentModule`** instead of agent-name switches.
> 
> **Session restore** gains a keyed **`RESUME_SUPPORT`** table and **`resumableAgents()`** driven by **`caps.resume`**, so a future resumable agent cannot fall through to Codex’s pointer/argv. **`fork`**, **`attach`**, cloud attach, and custody paths similarly derive from **`caps.resume`**, registry order, or **`agentIds()`** instead of hardcoded lists.
> 
> **Tests replace closed unions**: `no-inline-agent-union`, `agent-module-table`, `agent-command-coverage`, plus resume-support alignment with the registry. Hub/types and **`resolveAgentLauncher`** follow the same identity model (generic positional seed launcher). Provider SDK **2.11.0** widens published agent id typing without an API version bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bb56d0372bd00d657ef3e2ca9310dd7e308452e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Both findings confirmed and fixed in 1bb56d03 — same class, and one my own coverage guards missed: they check table *keys*, not per-agent *branching*.

**Resume treats non-Claude as Codex** — real. `agentResumeArgs`'s `else` was total while the parameter was `'claude' | 'codex'`; widening it to `AgentId` left a fourth `caps.resume` agent getting Codex's marker file, `resume --last` argv and — on docker — the `codex` binary via `startCodexSession`. No behavior change today (claude and codex are the only `caps.resume` agents), but a latent trap. The claude/codex specialness is now one `RESUME_SUPPORT` table holding all three per-agent decisions (pointer probe, skip-permissions, docker starter); an agent with no entry gets "nothing to resume" rather than someone else's session.

**Attach drops unknown agents silently** — real, and worse than latent-only in shape: the session probe already accepts any `agentIds()` name, so an unhandled agent fell out of the switch and `agentbox attach` exited 0 having attached to nothing. Now a keyed `ATTACH_WRAPPED` map that throws, matching `loadAgentModule` / `agentCommandFor`.

Both new maps are covered by the existing guards (`agent-command-coverage` gained the attach map, `agent-module-table` gained a resume-support assertion), and both were mutation-checked — dropping an arm from either fails.

bugbot run

madarco:
bugbot run

PR #343, opened 2026-08-30
feat(agents): one CLI command factory for every agent

madarco:
Item 3 of the agents-as-in-monorepo-plugins plan: `agentbox claude`, `agentbox codex` and `agentbox opencode` become one command built three times.

**Base is #342**, not `feat/custom_agents` — this builds directly on its module table. Merge #342 first, then this rebases onto `feat/custom_agents`.

## What it was

4,866 lines across three hand-maintained files for what is one command. Measured, not estimated: normalise the agent's name out of `codex.ts` and `opencode.ts` and `diff` is **671 lines** — ~1,070 were byte-identical. Claude was closer than its 2,062 lines suggested: normalised against codex it added exactly **four** options (`--plan`, `--headless`, `--code`, and one login flag); the extra ~600 lines were the headless-login machinery and the `--plan` / plugin-rebuild blocks woven through a body that was otherwise the same.

## What it is

- `agents/command/` — the option surface, the create body, `start`/`attach`, the default `login`, and the type contract between them.
- `agents/<id>/runtime.ts` — the agent's docker bindings and its own login code. Separate from the command so `agent-sessions.ts` can restart a session without pulling a commander tree behind it.
- `agents/<id>/command.ts` — a descriptor: the help strings that genuinely differ, and hooks.

Two tables under `agents/`, both with literal `import` specifiers: the lazy `AGENT_MODULES` (spec / login / teleport / runtime) and the eager `AGENT_COMMANDS` (the commander tree). `index.ts`, `attach.ts`, `fork.ts`, `create.ts`, `argv-prefix.ts` and `agent-sessions.ts` all read them — none of them names an agent any more.

**Per-agent behavior is a closed set of five hooks**, and claude is why they exist. A sixth would mean the body is not actually shared.

Three things that looked like accidents are now *declared*, because each has a reason: `signInOfferTiming` (claude asks before its wizard can spend minutes re-baking a stale base; the others ask after hub routing, so a hub-built box never prompts for a local login it won't use), `acceptsSeedPrompt: false` for opencode, and `ensureInstalledOnCreate: false` for claude.

## How the "changed nothing" claim is backed

`test/_fixtures/agent-cli-surface.json` was captured from the three hand-written commands **before** the factory existed — every flag, short form, description, default and positional argument of all three commands and their subcommands. It is unchanged.

It is also not sufficient, and that showed up the honest way. A live smoke caught the one regression a surface fixture structurally cannot: with teleport living in each agent's own hook, `agentbox opencode -c` — no teleport, therefore no hook — silently ignored the flag and **built a box** instead of refusing. The flag was still declared, so the fixture was green. Teleport now resolves in the shared body for every agent, and `agent-resume-flags.test.ts` drives that body directly.

## Verified

- Golden CLI surface unchanged for all three agents; confirmed the guard bites by editing one help string.
- Every new guard mutation-checked: dropped an agent from the command table, gave opencode a resume probe it doesn't declare, made the teleport refusal fall through.
- Packed install (`npm pack` → clean prefix): `claude --resume <uuid>` reaches the lazy agent module, so the literal-specifier rule holds in the published bundle, not just the dev tree.
- **Live docker, all three agents**: create → session up → `start` on a fresh session (config-volume sync, codex hooks seeded, claude's setup skill + plugin rebuild) → `agentbox list` shows `claude:idle` → `download claude --dry-run` → `attach`'s no-session path → destroy.
- Full CI locally: `check:plugin-skill`, `lint`, `build`, hub standalone, `check:cloud-backends`, `typecheck`, `test` (4,228 tests).

## Not in scope, and why

`commands/list.ts` still branches per agent. It reads the five named `BoxStatus` fields, which are a wire contract the hub `/api/v1` payload and the macOS tray consume — that is the ctl status-map item (5), not this one. The plan previously listed `list.ts` here and that was wrong.

Per-agent config keys survive too: `@agentbox/config` is a zero-internal-dep leaf (`sandbox-core` depends on *it*), so it cannot read the registry to generate `agents.<id>.*`. Recorded as its own item rather than smuggled in — it is a change to a documented, shipped surface.

---

> [!NOTE]
> **Medium Risk**
> Large CLI refactor with behavior-preservation claims; the shared create/teleport/resume paths and login flows are high-touch user entry points where subtle ordering or agent-registry gaps could regress.
> 
> **Overview**
> **Replaces three ~4,800-line duplicate `agentbox claude|codex|opencode` implementations** with `buildAgentCommand()` plus thin per-agent descriptors. Shared logic lives under `agents/command/` (options, create, start/attach, default login, types); agent-specific docker/login/session-resume behavior lives in `agents/<id>/runtime.ts` and is exposed on the lazy `AgentModule` so paths like `agent-sessions.ts` do not import Commander.
> 
> Each agent’s `command.ts` now only supplies differing help text, `AgentRuntime`, and up to five hooks (e.g. Claude `--plan`, setup wizard, plugin rebuild, clipboard paste). **`agent-sessions.ts` drops the hardcoded `RESUME_SUPPORT` table** and delegates resume/start/skip-permissions/session names to `runtime` + `caps.resume`. **`resolveResumeSeed` runs in the shared create body** so agents without teleport (OpenCode) refuse `-c`/`--resume` instead of silently creating a box. **`agentbox attach`** dispatches through `agents/commands.ts` instead of a local wrapper map.
> 
> Intentional per-agent flags are explicit on the spec: `signInOfferTiming`, `acceptsSeedPrompt`, `ensureInstalledOnCreate`, and cloud attach now uses configured session names via `runtime.sessionNameOf`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6623a868f6d64cd1918851eb2e1f22a9c226f8f2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Good catch — and it predates the factory. All three hand-written commands did `opts.sessionName ?? 'claude'|'codex'|'opencode'` on the cloud branch while the docker branch read `<agent>.sessionName`, so a custom session name meant create started one session and a later cloud attach silently created a second. The refactor preserved it faithfully (the golden CLI-surface fixture wouldn't see it — it's a value, not a flag).

Fixed in 78539035: one `resolveSessionName(a, opts, cfg)` used by both cloud branches, with `agent-session-name.test.ts` pinning both halves — explicit `--session-name` wins, otherwise the configured name and not the registry default. Verified the test bites by reverting the helper to `a.spec.sessionName`.

This is the one deliberate behavior change in the series; everything else is held byte-identical by the fixture.

madarco:
bugbot run

madarco:
bugbot run

madarco:
Right, and this one is mine — not pre-existing. The hand-written command printed a literal there; I turned the reason into an optional field on the preflight result that a hook could supply, and claude's `--plan`-only path set `hubIncompatible: true` without it. So `agentbox claude --plan … --via-hub` logged a blank line and built locally with no explanation. Invisible to the CLI-surface fixture, since no flag changed.

Fixed structurally in 6623a868 rather than by filling the field in: `AgentPreflight.hubIncompatibleReason` is deleted and the body always reads `text.hubIncompatibleReason`, which `AgentCommandText` requires. There is no longer a place to forget it, so no test is needed to remember.

madarco:
bugbot run

PR #344, opened 2026-08-30
feat(agents): declarative hook seeding, and make it reach cloud boxes

madarco:
Item 4 of the agents-as-plugins plan.

## What

`seedSetupSkillIntoVolume` / `seedCodexHooks` / `seedOpencodePlugin` were the same throwaway-container copy written three times, differing only in which file went where. They become `AgentSyncSpec.seeds` — baked source, destination relative to the config root, and the `runtime/_shared` basename to fall back to — plus `launchFlags` for the argv an agent needs to *load* what was placed.

## The bug this fixes

Living in `@agentbox/sandbox-docker` meant **the cloud providers never ran any of it**:

- OpenCode's state plugin is its only activity source (`caps.activitySource: 'plugin'`) and was not even shipped to the VPS providers — so **every cloud OpenCode box reported `unknown` activity forever**.
- Codex never got `~/.codex/hooks.json` on cloud, nor the flags to load it (the cloud launcher ran the bare binary).

Same drift class as `boxRunEnv` before it was wired. Both transports now run one plan built in `sandbox-core`, asserted equal key for key.

The cloud path prefers the baked copy and falls back to uploading the host's staged `runtime/_shared/` copy — which is what lets this land on **existing** snapshots. The deployed Hetzner base has no OpenCode plugin to copy, so without the fallback a re-`prepare` of every provider would have been a prerequisite.

## Verified live

- **Docker**, all three agents: files land in the config volume, vscode-owned, every intermediate dir writable by the box user; codex's launch argv carries `--enable hooks --dangerously-bypass-hook-trust` from the spec.
- **Hetzner (real VPS)**: `seeded the agentbox-state plugin (from host)` — the upload fallback firing on a snapshot that predates the asset — with the plugin present and correct in the box. Deleted it, ran `opencode start`, watched it come back. Server + firewall cleaned up after.

## Two things the tests could not have caught

1. **Seeding on `attach` but not on `start`.** The first version put the cloud call on the CLI action closures, and `wireAttachAction` / `wireStartAction` are separate bodies. Fixed structurally — the call moved onto `cloudAgentAttach` / `cloudAgentStartDetached`, the two primitives every cloud launch funnels through — and pinned by a test that fails if either drops it or seeds after the launch.
2. **A spinner frame is not evidence.** The create-time seed log line never rendered for OpenCode (set and replaced inside one frame) while codex's happened to survive. The file in the box was the ground truth.

## Tests

New `agent-seed.test.ts` (the `boxRunEnv`-style docker == cloud drift assertion, the invariant a plugin-activity agent must declare its seeds, and that both sources really carry each asset), plus `cloud-seed-launch.test.ts`. Eleven mutations verified to bite. Full CI green locally.

---

> [!NOTE]
> **Medium Risk**
> Changes agent launch and config seeding on every docker start/attach and all cloud attach/detached paths; failures are best-effort but missing seeds still degrades activity reporting until the next successful seed.
> 
> **Overview**
> Replaces three docker-only seed helpers with **`AgentSyncSpec.seeds`** and **`launchFlags`** on the agent registry, plus shared planning/script building in `sandbox-core` so docker and cloud run the same placement plan.
> 
> **Docker/local CLI:** Per-agent `afterVolumeSync` / dashboard one-offs go away; **`seedAgentDeclaredFiles`** runs from the shared start/attach path (and create-time docker/cloud sync) for Claude’s setup skill, Codex `hooks.json`, and OpenCode’s state plugin. Codex docker launches now read **`launchFlags`** from the spec instead of a local constant.
> 
> **Cloud (the bug fix):** Adds **`seedAgentDeclaredFilesViaTransport`** with baked-image copy plus **`runtime/_shared`** upload fallback for older snapshots. Seeding runs on box create (`seedAgentConfig`), after resume in **`cloudAgentAttach`**, and inside **`startDetachedCloudAgent`** after `provider.start` (start → seed → launch). **`buildCloudAttachInnerCommand`** prepends declared **`launchFlags`** before user args so Codex can actually load seeded hooks on cloud attach/resume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6dbce8b4b41cf3f738ef5a926723af073bf9c83. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #345, opened 2026-08-30
feat(status): one keyed agent status map, and make a fourth agent able to report

madarco:
Item 5 of the in-repo-agents series. `BoxStatus` carried a required `claude` plus
optional `codex`/`opencode`, three ctl ops, three reporter setters and three
snapshot branches; `list`, the dashboard, the attach footer and the relay's queue
gate each re-spelled the same three names. **A fourth agent had nowhere to put an
activity value** — not "wasn't wired up", but had no field.

Status is now a keyed `AgentStatusMap` in `@agentbox/core` (the only leaf both ctl
and the relay can reach — ctl depends on the relay, not the reverse), behind one
`agent-state <agent> <state>` op. ctl learns which tmux sessions to probe from the
`agents.list` descriptor it already fetches, so **an agent added after an image was
baked is probed without a re-bake** — the `CredentialsWatcher.setFiles()` shape,
started on the baked list and upgraded detached, never on the daemon's critical path.

## Three live bugs, all "the field simply wasn't there"

1. **OpenCode activity was produced, transported, persisted, then dropped.**
   `opencodeActivity` existed nowhere in the repo; `lifecycle.ts` projected only the
   session title, so `list` could print a bare `opencode` and nothing more, and an
   OpenCode box that errored couldn't raise the box status to `error`. #344 had just
   finished making that plugin reach cloud boxes and the result was invisible.
2. **`agent state` / `wait-for` / `get-plan-question` were claude-only** —
   structurally: `getAgentState` returned `{ claude }` and every wait predicate took
   `BoxStatusClaude`, so on a codex or opencode box they read claude's permanent
   `unknown` forever. They now read the box's most active agent, or `--agent <id>`.
3. **The hosted Postgres plane** emitted none of the five per-agent fields, unlike
   the in-process one, so `list`'s AGENT column silently degraded against a control box.

## Two things deliberately unchanged, both load-bearing

- **`schema` stays `1`.** `readBoxStatus` and the relay's `isValidBoxStatus` both
  reject anything else outright, so a bump would blank every field on every existing
  reader rather than degrade. `agents` is additive — the discipline `probed`,
  `expose` and `sessionTitle` already follow — and back-compat is read-time
  normalization, the rule `normalizeLastAgent` states for persisted agent names.
- **The named fields are still written and published.** The snapshot writes them as
  a *derived* mirror (a test asserts it moves with the map; mutating that took down
  11 tests). Skew runs both ways — baked ctl outlives the host that reads it, and a
  laptop CLI is routinely a different build from the control box it points at — and
  those fields drive the queue gate, autopause and keepalive, not just display. The
  macOS tray decodes the three `*SessionTitle` keys **by name as optional strings**,
  so removing one would blank box labels silently rather than fail.

The seeded hook files still call `<agent>-state` and ctl still ships those names
(generated from the built-in list): agent config volumes are **shared between boxes**,
so a `hooks.json` seeded by a newer image can be read by a box running older ctl.

## Verification

Live on docker (all three agents) and on a real Hetzner VPS, all boxes destroyed and
no cloud orphans.

- `opencode:working` renders in `list` on docker **and** hetzner — the headline bug.
- `agent state` returns `waiting` on an opencode box; `wait-for idle` resolves.
- Claude with real hooks: `idle → working → idle`, `updatedAt` advancing, session
  pointer still captured by `--capture-session`, mirror tracking the map field-for-field.
- **The back-compat proof landed better than planned:** a codex box built from the
  current image turned out to run *genuinely pre-change ctl* (no `agent-state`
  command), so it posted the old named blocks for real — and the new host rendered
  `codex:idle`, with the relay deriving `agents` at ingestion. Same through the
  Hetzner *poller* door. That is the exact skew the design exists for, observed
  rather than simulated, on both ingestion paths. A checked-in old-shape fixture
  covers the third door (`readBoxStatus`).
- An agent id nobody has heard of (`openclaw`) renders end to end.

Full CI locally: lint, build, standalone, cloud-backends, typecheck, 4,243 tests.

## Also in here

`notify` was agent-agnostic in name only; box status could be raised to `error` only
by claude or codex; `caps.activitySource` went from zero production reads to a list
(`hooks`/`plugin`/`scraper`) with three honest consumers. The last commit fixes the
add-an-agent checklist in `docs/agents.md`, which still described the surface this PR
removes.

---

> [!NOTE]
> **Medium Risk**
> Broad changes to the status snapshot, hub list/agent APIs, and orchestration (`wait-for`), but mitigated by read-time normalization and continued legacy field publishing across version skew.
> 
> **Overview**
> Replaces the fixed **claude / codex / opencode** status fields with a keyed **`agents` map** (`AgentStatusMap` in `@agentbox/core`), so any agent id can report activity and consumers iterate the map instead of hardcoding three names.
> 
> **CLI & hub:** `agent state`, `wait-for`, and `get-plan-question` gain **`--agent <id>`** and default to the box’s **most active** agent via `pickPrimaryAgent` / `normalizeAgentStatus`. `list`, dashboard, cmux, and attach footers read **`agentStatus`** (with `hubBoxAgentStatus` falling back to legacy named fields for older hubs). **`GET /boxes/{id}/agent`** returns `{ agents, claude }`; list payloads add **`agentStatus`** and **`opencodeActivity`**.
> 
> **ctl:** Three separate state commands collapse behind **`agent-state <agent> <state>`** (legacy `<agent>-state` names kept for shared hook volumes). The daemon upgrades **which tmux sessions to probe** from **`agents.list`** without re-baking the image.
> 
> **Back-compat:** `BoxStatus.schema` stays **1**; old snapshots normalize on read and current writers still mirror the three legacy blocks for trays, queues, and mixed CLI/hub versions.
> 
> Fixes **OpenCode activity never surfacing**, **agent CLI commands always reading claude’s `unknown` on codex/opencode boxes**, and **hosted Postgres list rows missing agent columns**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63e25970b33e9a097fa83905887ec87f0fedb519. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #347, opened 2026-08-31
Phase 2a: each agent owns its login binding, unblocking the CLI-layer move

madarco:
Follow-on to #346. Starts phase 2 (the CLI layer) by removing what was actually blocking it.

## The blocker was an import edge, not a size

The plan recorded phase 2 as blocked on "26 distinct shared CLI modules". That count is right but it was the wrong thing to measure. Re-measuring changed the plan:

- **All six HEAVY imports belong to claude alone** — `commands/prepare` (975 lines), `commands/control-plane` (2,729), `wizard.ts`, `provider/registry`, `checkpoint-lookup`, `commands/_errors`, every one from `agents/claude/{command,login-command}.ts`. **codex (606 lines) and opencode (371) touch none of them.** They need ten modules, and each is a near-leaf (0–4 local imports of its own).
- **What genuinely blocked everything** was `lib/agent-login-bindings.ts`: it held three `<agent>LoginBinding` functions and imported each agent's login spec and docker helpers — a per-agent table in shared code, pointing from the shared CLI *into* the agents.

## The fix

Each of those three had exactly one caller: that agent's own runtime, which already exposes the binding through `AgentRuntime.loginBinding`. The seam was already there; the functions were on the wrong side of it. They now live in `agents/<id>/login-binding.ts`, and the shared file keeps only the `AgentLoginBinding` contract and `withLoginDefaults` — which is all `guided-login.ts` and the command descriptor ever used, both by type only.

## The guard had a hole, and this is what fell through it

`no-agent-named-exports` only matched capitalized agent names, so `claudeLoginBinding` sat in the shared CLI unnoticed. Broadened to the lowercase form in the same pass; six more files came into view, each allowlisted against its owning phase.

One is excluded rather than exempted: `commands/_open-in.ts`'s `codexAddUrl` is the **Codex.app** `codex://` deep link — the desktop app, sitting among herdr, cmux, vscode and finder. Renaming it would be wrong, which is the same carve-out `no-inline-agent-union.test.ts` already makes for that file.

## Verification

Full CI locally (`check:plugin-skill`, `lint`, `build`, hub standalone, `check:cloud-backends`, `typecheck`, `test`). New tests assert each binding mounts its **own** config volume and never another agent's, that only claude carries post-success work, and that a user's args beat the agent's defaults — all three mutation-checked, plus the broadened guard.

## Next

Extract the ten-module kit and move codex + opencode; then claude's heavy six, most likely inverted through the existing `AgentCliSpec` ctx seam rather than by extracting `control-plane.ts`.

---

> [!NOTE]
> **Low Risk**
> Structural refactor with behavior-preserving tests; login flow logic is moved, not rewritten.
> 
> **Overview**
> **Phase 2a** removes the import edge that blocked moving agent CLI code into packages: shared `lib/agent-login-bindings.ts` no longer imports each agent’s login spec and docker helpers.
> 
> **`claudeLoginBinding`**, **`codexLoginBinding`**, and **`opencodeLoginBinding`** now live in `agents/<id>/login-binding.ts` beside each runtime; runtimes import locally instead of from the shared lib. The shared module keeps only the **`AgentLoginBinding`** contract and exported **`withLoginDefaults`** (renamed from internal `withDefaults`) for **`guided-login.ts`** and command types.
> 
> **`no-agent-named-exports`** now flags lowercase agent-prefixed exports (e.g. `claudeLoginBinding`); six additional files are allowlisted for later phases, and **`commands/_open-in.ts`** is excluded because **`codexAddUrl`** names the Codex desktop app deep link.
> 
> New **`login-binding.test.ts`** asserts behavior-preserving mounts, Claude-only **`finalize`**, and default-args handling. **`docs/agents-as-packages-plan.md`** records 2a done and 2b next (kit extract + codex/opencode move).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 980fe9134a5384ad1ae8ed162f468665ba653862. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/agents/claude/host-creds.ts
+++ b/apps/cli/src/agents/claude/host-creds.ts
@@ -1,0 +1,52 @@
+/**
+ * Claude's host-side credential check, for the pre-flight that runs before a
+ * detached `-i` job. Beside the runtime that owns it — the shared queue helper
+ * used to hold this and codex's and opencode's, dispatching between them with a
+ * three-arm chain that a fourth agent fell through silently.
+ */
+import { hostBackupHasCredentials } from '@agentbox/sandbox-docker';
+import { resolveClaudeAuth } from '../../auth.js';
+import { resolveClaudeCredHealth } from '../../lib/claude-cred-health.js';
+
+/**
+ * True when Claude is already authenticated on the host: a forwarded env var
+ * (`ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN`), the legacy
+ * `~/.agentbox/auth.json` setup-token, or a real OAuth refresh token in the
+ * host backup (`~/.agentbox/claude-credentials.json`). The backup is what the
+ * foreground sync writes whenever a box's claude logs in, so its presence is
+ * the load-bearing signal that the shared volume has been seeded.
+ */
+export async function claudeAuthAvailable(env: NodeJS.ProcessEnv): Promise<boolean> {
+  const resolved = await resolveClaudeAuth(env);
+  if (resolved.source !== 'none') return true;
+  return hostBackupHasCredentials();
+}
+
+/**
+ * Richer Claude credential verdict for the non-interactive paths. `'missing'`
+ * when nothing can seed a box; `'expired'` when the saved login can no longer be
+ * renewed AND we're on a cloud provider — cloud has no shared volume to fall
+ * back to, whereas a docker box boots from the volume's live copy and refreshes
+ * it in-box. A host-env token (`ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN`)
+ * or legacy `auth.json` short-circuits to `'ok'`: it has no expiry concept here.
+ * Mirrors the interactive `maybeRunCloudClaudeLogin` split, and shares its
+ * verdict helper — so a merely-lapsed access token is renewed here too rather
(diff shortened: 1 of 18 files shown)


madarco:
bugbot run

madarco:
Pushed 237b5360 — a real bug found while sizing the phase-2 kit, not by a test.

`assertAgentCredsAvailable` dispatched with `agent === 'codex' ? codexAuthAvailable : opencodeAuthAvailable`, so **any agent that was neither claude nor codex was checked against OpenCode's env keys and volume** — a wrong answer with nothing failing. Each agent's host check now lives beside its runtime, and `AgentRuntime.hostCredStatus` is a required member, so an unwired agent is a compile error rather than a silent pass.

That emptied `assert-creds.ts` of agent names, and the guard's stale-exemption half then failed demanding its removal from the allowlist — which is the half that makes the list shrink rather than accumulate. 27 files now, down from 29.

bugbot run

madarco:
Merged into `feat/custom_agents` (fast-forward, `237b5360`).

To be precise about coverage: Bugbot's clean review was of `980fe913` (phase 2a). It did not re-review `237b5360` (the credential-dispatch fix) despite two `bugbot run` requests — the same non-re-review pattern as #345 and #346. That commit went in on green CI plus four mutation-checked regression tests and a live check that all three agents answer through their own `hostCredStatus` against real host state.

PR #349, opened 2026-08-31
Agents as packages: phases 2c, 7 and 4 (claude's descriptor, ctl renames, SDK v3)

madarco:
Phases 7 and 4 of the agents-as-packages plan. Continues #348.

## Phase 7 — the last ctl/core renames

`claude-session` → `agent-session <agent>`, and the socket op with it. Safe because — unlike the `<agent>-state` commands, which are frozen since seeded hook files in *shared* config volumes invoke them by name — this one has **no caller at all**: not a seeded file, not the host (which probes tmux directly through the registry's `sessionInfo`), not another package.

`session-pointer.ts` had four per-agent functions, and ctl held **two** hand-written tables keyed by agent, each carrying half the same fact. Now one table with each agent's pointer path and shape (`session-id` for claude, `presence` for codex), and three generic functions.

**The path values do not move, and now cannot.** The writer is the ctl baked into a box's image; the reader is the host CLI, which greps the literal strings. Renaming one silently costs every existing box its resume-on-restart. A test pins them — hardcoded rather than derived, because deriving them would make it agree with any rename.

That test caught a hole worth having: nothing stopped `markAgentActive` writing a timestamp over claude's session id, and `status-reporter` calls it for every agent on first activity.

**One deviation from the plan:** `core/claude-tui.ts` is re-tagged to phase 4 rather than renamed. Reading it changed the verdict — it returns Claude Code's *own* env vars and exists because of Claude's renderer, so `agentTuiEnv` would claim a generality it does not have. The real leak is `binary === 'claude' ? claudeTuiEnv(...)` in `sandbox-cloud/src/detached-agent.ts:337`, which wants spec data, not a rename.

## Phase 4a — SDK v3 (breaking)

`SUPPORTED_SDK_API_VERSIONS = [3]`, per your call for a clean break.

v2 exported three per-agent stagers, so a provider built on it bakes a base containing exactly claude/codex/opencode — on a host with a fourth it produces a snapshot silently missing it. That is a *wrong box*, not an absent feature, which is why the older majors are refused rather than tolerated. v3 exports `stageAllAgentStatic` + `AgentStaticStage` instead.

Both examples move to `providerApiVersion: 3` (that field is load-bearing — `plugin add` reads package.json **before** the SDK's own constant), and the reference example is ported off its hand-rolled three-block staging.

**Adds the CI that would have caught this.** `examples/` is outside `pnpm-workspace.yaml`, so nothing built, typechecked or tested it — this bump could have refused both examples with the suite green, which is exactly how the agent example shipped broken in #348. The new test also pins the two halves of the version decision that live in different packages: the SDK's `SDK_API_VERSION` against the CLI's supported set, and the package major against both. `pack:test` only validated the SDK side.

## Phase 4c — a second hardcoded triple, and the bug in it

`sandbox-cloud/src/sync/agent-credentials.ts` declared its own three-row `AGENT_SPECS`, and every consumer filtered against it — so **a fourth agent got no credentials mount, no static mount and no seed in a cloud box**. All three fields were already on the spec, exactly; the rows derive now, and the staging closures arrive through the existing `AgentCloudModule` seam (claude had no cloud module and now registers one).

Also fixes a latent data bug it exposed: `credential.cloudSubpath` is documented as a sub-*dir* and every built-in uses `<id>/`, but the example spec said `auth.json`. Fixed in the registry row and the shipped example, and `agentSpecProblem` now rejects it — the three earlier validation passes did not.

## Verification

Full CI locally at every commit, plus `pack:test` on the SDK (`OK — 31 exports, SDK_API_VERSION=3`). Every new guard mutation-checked in both directions. Allowlist 27 → 24.

**Not yet run: a real cloud `prepare`.** It bakes a snapshot, and it is the check I most want before this is trusted — the `hostHomeRel` bug in #348 lived in exactly this staging path. Doing it next.

---

> [!NOTE]
> **Medium Risk**
> Touches the shared agent create/login/attach pipeline and introduces SDK v3 breaking changes for provider plugins; behavior is largely preserved via injection but regressions would show up at box create, login, and cloud bake time.
> 
> **Overview**
> This PR finishes **agents-as-packages** for the CLI: shared types, prompts, progress, teleport helpers, and `defineAgentModule` move into **`@agentbox/cli-kit`**, while **Claude/Codex/OpenCode descriptors** (wording, hooks, runtimes, login) live in each **`@agentbox/agent-*/cli`** package. **`apps/cli` keeps only ~15-line shims** that call `buildAgentCommand(<spec>)` and the shared create/attach factory.
> 
> **Dependency inversion** is what unblocks Claude’s large descriptor: hooks no longer import the app. **`AgentHostServices`** on the create context (wired in **`host-services.ts`**) supplies setup wizard + optional base re-bake, plan staging, install hints, and clipboard paste; **`attachExtras`** receives a **`clipboard`** bag; custom **`login`** gets injected **`syncCredentialsIfChanged`** / **`fail`**.
> 
> The **demo `example` agent** is registered end-to-end (module table + command) but **hidden from `--help`** via registry `hidden` flags; coverage tests now expect **every registry agent**, not just visible ones.
> 
> Related doc/example updates in the diff include **provider SDK v3** (`stageAllAgentStatic`, `providerApiVersion: 3`), **cloud attach** using registry-driven **`agentTuiEnv`**, wizard discriminant **`command: 'agent'`**, and a **provider-examples CI test** so out-of-workspace examples cannot drift from the CLI’s supported API version.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99f31a4393bb2c319c5f7a55685999bdc287033b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
Two more commits since that review — phase 2b's foundations:

- `@agentbox/cli-kit` extracted (12 files, ~1,050 lines, 55 importers rewritten). Worth flagging: I first named it `packages/agent-cli-kit`, which put it inside the `packages/agent-*` glob the no-agent-named-exports guard *exempts* — silently making everything in it invisible. A planted `claudeSpinnerHack` export passed the suite. Renamed to `cli-kit`, added to the guard's roots, and the same mutation now fails.
- The agent contract (`AgentRuntime`/`AgentCliSpec`/`AgentModule`) moved into the kit, so an agent package can implement it. `CreateRouting`, `HostCredVerdict` and `AttachExtras` came with it; the app now re-exports rather than redeclaring.

bugbot run

madarco:
Phase 2 is complete — `apps/cli/test/_agents-in-cli.ts` is deleted, which was the phase's own stated proof.

The demo agent found one more real bug on the way: `agentbox example` produced a box with **no config volume at all**, because `create.ts` still resolved three hand-written `<agent>Spec` consts. Both the specs and the mounts derive from the registry now; verified live that the demo agent gets `agentbox-example-config` mounted and that a claude box built right after still gets its own.

Eight commits since your last review — a fresh pass would be valuable, particularly on `packages/sandbox-docker/src/create.ts` (the daily-driver path) and the CLI-kit extraction.

bugbot run

madarco:
Phase 4b is complete — `host-stage.ts` holds no agent's staging any more. Two things worth a reviewer's eye:

- `seedClaudeJsonAtCreate` was still called by name from `cloud-sync.ts`. The open question was never whether to invert it but whether inverting would **move it in the sequence** (it must run after the declared seeds and needs `hostWorkspace`), so it got its own `afterDeclaredSeeds` hook at exactly that point rather than being folded into `afterSeed`. The existing order test proves the position — mutating it earlier fails.
- I briefly introduced `Circular package dependency: agent-codex, agent-claude, sandbox-cloud` by importing the real stagers into a `sandbox-cloud` test. Backed out; that test stages with `sandbox-core` primitives now, and codex's Keychain case moved into its own package.

Verified against the real host home: same five stages dispatch, claude's hook-filtered `settings.json` is 31,463 bytes — byte-identical to before the move.

bugbot run

madarco:
@cursor review

madarco:
Fixed in 41140835.

`startExampleSession`, `exampleSessionInfo` and `buildExampleAttachArgv` now take the caller's session name and args instead of always using the spec default. `cli/runtime.ts` forwards all three.

Added `packages/agent-example/test/session-params.test.ts` (3 tests). It reaches `sessionInfo` through the package function rather than the runtime adapter on purpose: the CLI contract narrows that return to `{ running: boolean }`, so a dropped session name is invisible through the adapter's type. Mutation-checked — reverting to `SPEC.sessionName` fails the test.

madarco:
Pushed four more commits since the last review:

- `dda3794e` **phase 2c** — claude's command descriptor moved into `@agentbox/agent-claude` (289 lines -> a 22-line shim). Its hooks called UP into the app (`runPrepare`, the setup wizard, `providerForBox`, the clipboard helpers), which would have closed an `apps/cli -> package -> apps/cli` cycle; they are passed in now via `AgentHostServices` and a `clipboard` argument to `attachExtras`. Three of the six collapsed into one `host.setupWizard()` — claude never *read* `baseStatus`, it only forwarded it.
- `44633650` **phase 7** — ctl's three `DEFAULT_<AGENT>_SESSION_NAME` constants deleted. Removing them surfaced that the `agent-session` socket op read `req.sessionName ?? DEFAULT_CLAUDE_SESSION_NAME`, so a caller omitting the field got a probe of *claude's* session and a confident answer about it. Now fails closed, matching its sibling `agent-state`.
- `e985fba9` **phase 4b** — opencode's host stagers moved into its package, unblocked by the v3 SDK dropping the per-agent staging exports.
- `980e82eb` **phase 4** — the TUI renderer pin is `AgentSyncSpec.tuiEnv` data now; all three `binary === 'claude'` branches deleted.
- `99f31a43` **fix** — `fetchWatchList()` sat inside the credential-sync gate, so `--no-credential-sync` boxes never upgraded their agent *probe* list either; a plugin agent was invisible in `agentbox list` there, silently.

Allowlist in `no-agent-named-exports.test.ts`: **27 -> 16**. The guard's blanket `apps/cli/src/agents/` exclusion is narrowed to the four command shims, so the create/attach pipeline is now measured too.

Verified live on docker (not just typechecked): create with a working claude session; `--plan` failing fast on a bad path and landing a real plan with `--permission-mode plan` in the argv; the ctl socket refusing a session-less probe in a rebuilt image; a box driven by the fourth (example) agent; and a packed install in a clean prefix. Every new guard mutation-checked.

madarco:
bugbot run

PR #351, opened 2026-08-31
Phase 4b: claude's credential guards, the renewal, and the staged-item mappers

madarco:
Continues the agents-as-packages work. Allowlist in `no-agent-named-exports.test.ts`: **16 -> 13**.

### `64327f94` — credential guards + renewal into `agent-claude`

`hostClaudeAccessTokenExpired` / `hostClaudeLoginDead` (sandbox-core) and `renewClaudeCredential` (sandbox-docker) only mean anything for a `claude-oauth` blob, and their only consumer was already `@agentbox/agent-claude`. Moving the renewal also deletes an inversion: it reached claude's own warm-up through `requireAgentSyncModule('claude')` purely because it sat in the wrong package.

The OAuth field parsers stayed in `sandbox-core`, renamed for the shape rather than the agent (`oauthExpiresAt` / `oauthRefreshExpiresAt`) — the generic accept-gate needs them and dispatches on `credential.realShape`.

**Defect found on the way:** `dockerCredentialRefresh` named three agents in sequence — claude sync gated on claude's expiry, then codex extract, then opencode extract. A fourth agent got **no host-backup refresh at all**, silently (every step is best-effort), so its cloud boxes were seeded from a stale token. It walks the module registry now via a new optional `AgentSyncModule.refreshHostBackup`; verified live that all three agents register one, so coverage is unchanged.

`resolveClaudeCredHealth` gained an explicit `probes` seam — those four IO calls were mockable before only because they crossed a package boundary.

### `601902a9` — staged-item mappers into their packages

`claudeStagedItems` / `codexStagedItems` / `opencodeStagedItems` were three per-agent functions in the agent-neutral propagate layer. Each moved to `packages/agent-<id>/src/staged-items.ts` with its test; the `StagedItem` contract and transport stay shared.

### Verification

Full gate green (plugin-skill, lint, build, hub standalone, cloud-backends, typecheck, test). Every new guard mutation-checked — narrowing the registry walk back to one agent fails the test. Live on docker: box create, and the box's claude completing a **real authenticated turn** (`claude -p`), which is the check that actually exercises the credential path end to end.

---

> [!NOTE]
> **Medium Risk**
> Touches pre-box credential gating, cloud seed refresh, and OAuth renewal paths; behavior is intended to be preserved but registry-driven refresh and moved guards affect every box create on docker/cloud hosts.
> 
> **Overview**
> **Phase 4b** of agents-as-packages: Claude-specific credential logic and per-agent download propagate mappers leave shared `sandbox-core` / `sandbox-docker`, and host backup refresh becomes registry-driven so new agents are not silently skipped.
> 
> Claude OAuth **access vs refresh** health (`hostClaudeAccessTokenExpired`, `hostClaudeLoginDead`) and **`renewClaudeCredential`** move into `@agentbox/agent-claude`, with `resolveClaudeCredHealth` exposing an injectable **`probes`** seam for tests. Generic OAuth field helpers in core are renamed to **`oauthExpiresAt` / `oauthRefreshExpiresAt`**. CLI **`download-*`** commands import **`claudeStagedItems` / `codexStagedItems` / `opencodeStagedItems`** from each agent package instead of `sandbox-core`.
> 
> **`dockerCredentialRefresh`** no longer hardcodes claude → codex → opencode; it loops **`registeredAgentSyncModules()`** and calls optional **`AgentSyncModule.refreshHostBackup`** (Claude gates on access-token expiry; codex/opencode extract-only). **`renewClaudeCredential`** is no longer exported from `sandbox-docker`.
> 
> The three **`<agent>StagedItems`** mappers and their tests relocate to **`packages/agent-*/src/staged-items.ts`**; shared **`StagedItem`** / propagate transport stay in core. The **no-agent-named-exports** allowlist shrinks accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 601902a9d85c056724bfbeeac3ef8d20f076df28. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/src/commands/download-claude.ts
+++ b/apps/cli/src/commands/download-claude.ts
@@ -8,7 +8,6 @@ import {
 } from '@agentbox/agent-claude';
 import {
   agentBoxConfigDir,
-  claudeStagedItems,
   pullClaudeExtrasViaTransport,
   stageItemsViaTransport,
   type PullClaudeResult,
@@ -18,6 +17,7 @@ import { resolveBoxOrExit } from '../box-ref.js';
 import { cloudTransportForPull } from './_agent-pull.js';
 import { parsePropagateFlag, runPropagateStep } from './_agent-propagate.js';
 import { handleLifecycleError } from './_errors.js';
+import { claudeStagedItems } from '@agentbox/agent-claude';
 
 interface DownloadClaudeOpts {
   yes?: boolean;
--- a/apps/cli/src/commands/download-codex.ts
+++ b/apps/cli/src/commands/download-codex.ts
@@ -4,7 +4,6 @@ import { DEFAULT_BOX_IMAGE, stageItemsFromVolume } from '@agentbox/sandbox-docke
 import { pullCodexConfig, resolveCodexVolume, SHARED_CODEX_VOLUME } from '@agentbox/agent-codex';
 import {
   agentBoxConfigDir,
-  codexStagedItems,
   pullCodexConfigViaTransport,
   stageItemsViaTransport,
 } from '@agentbox/sandbox-core';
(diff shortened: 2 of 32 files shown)


madarco:
bugbot run

PR #352, opened 2026-08-31
Phase 2 remainder: codex's plugin installer and claude's path encodings out of the app

madarco:
Allowlist in `no-agent-named-exports.test.ts`: **13 -> 6**.

### codex's plugin installer
`apps/cli/src/commands/install-codex.ts` was 464 lines with ten exports, all about codex's plugin marketplace and its `~/.codex/config.toml` enable table. It is `@agentbox/agent-codex/cli` now; the command file is a 47-line shim.

`resolveDevRepoRoot` is **injected** — locating the user's source checkout reads the running CLI's own path, which is the app's business. Threading a parameter is where things get dropped, so there's a test that fails when `installCodexPlugin` forwards only `noDev` (the dev marketplace and skill symlinks would silently stop, with every build green). Verified live: `install codex --dry-run` still resolves the local repo root rather than the GitHub slug.

### claude's path encodings
`encodeClaudeProjectsDir` / `BOX_WORKSPACE_ENCODED` left `@agentbox/cli-kit`; `BOX_CLAUDE_PLANS_DIR` left `session-teleport/plan.ts`, which now takes `boxParentDir` and is agent-neutral.

### The bundling trap
Both moves invited importing the constant back from `@agentbox/agent-claude/cli` — and both times that breaks the **hub** build, not this one: `plan.ts` and `fork.ts` are bundled into the hub, and that entry drags cli-kit's pty backend in, so esbuild fails on a `.node` binary. The same bit `agent-codex`'s main entry, which `sandbox-docker`'s module registry pulls into the hub — hence the `./cli` subpath, plus a source-level test on the main entry.

### The guard got narrower
Rather than exempting the per-agent command files wholesale, it exempts the single `export const <agent>Command = new Command(` line. Those are correct (the dispatch tables need literal specifiers), but an agent's *logic* can no longer hide beside one — which is what forced the codex installer out. Planting a helper in `download-claude.ts` fails it.

### Verification
Full gate green (plugin-skill, lint, build, hub standalone, cloud-backends, typecheck, test). Every new guard mutation-checked. All four agent commands plus `install codex`, `fork` and `download <agent>` smoke-tested against the built CLI.

---

> [!NOTE]
> **Low Risk**
> Mostly structural moves with regression tests; behavior should match prior install/teleport/fork flows, with layering constraints documented to avoid hub bundle breaks.
> 
> **Overview**
> Continues **phase 2** of agents-as-packages: agent-specific logic leaves `apps/cli` and shared kit, while hub-safe imports are preserved.
> 
> **Codex plugin install** — The ~400-line `install-codex` implementation moves to `@agentbox/agent-codex` (`install-plugin.ts`), exported only from the **`/cli` subpath** so the hub bundle does not pull `@agentbox/cli-kit` (pty `.node`). The CLI command becomes a thin shim that **injects** `resolveDevRepoRoot` for dev marketplace + skill symlinks; tests guard against dropping that wiring.
> 
> **Claude paths** — `encodeClaudeProjectsDir`, `BOX_WORKSPACE_ENCODED`, and `BOX_CLAUDE_PLANS_DIR` move to `@agentbox/agent-claude`. They are removed from `@agentbox/cli-kit`. Generic `resolvePlanTeleport` now takes **`boxParentDir`** (claude passes its plans dir via `AgentPlanFileRequest`). **`fork`** uses `encodeClaudeProjectsKey` from **`sandbox-core`** instead of agent-claude/cli so hub bundling stays light.
> 
> **Guard** — `no-agent-named-exports` allowlist shrinks; only the single `export const *Command = new Command(` line is exempt per agent command file, so installer logic cannot hide behind command shims.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ab7f5f9012987475133543597a7bc3597253a1d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
Both findings fixed in `db65bc06` — Bugbot was right, and this had already bitten once inside `install-plugin.ts` before regressing again at the wizard's call.

Rather than just adding the argument, I removed the shape that allowed the drop: `resolveDevRepoRoot` is now **required** on `InstallCodexOptions` and `marketplaceSource`, so omitting it is a compile error and a caller with no checkout passes `() => null` explicitly. Verified by deleting it from the wizard again — tsc fails.

Worth noting the test I'd written covered the forwarding *inside* `installCodexPlugin` and still missed the second caller, which is why this needed a type change rather than more coverage.

PR #355, opened 2026-09-01
Add Pi (pi.dev) as the fourth agent

madarco:
Adds **Pi** ([pi.dev](https://pi.dev), `@earendil-works/pi-coding-agent`) as a first-class visible agent on every provider, with full session teleport, activity reporting and `agentbox download pi`.

Pi is an unusually good fit for a box: it ships **no permission prompts at all** — by design it runs with full tool permissions and its own docs recommend a container.

## What's here

- **Registry row + package** (`packages/agent-pi`), config row, and the two literal table arms. The install seam needed nothing: one `install` recipe drives the docker derived layer and every cloud derived snapshot.
- **Full teleport** (`caps.teleport: 'full'`). Pi stores sessions as per-cwd JSONL whose first line carries the `cwd`, so the resolver rewrites exactly one field and forwards `--session <box path>`. Verified live that Pi loads an out-of-tree session path and appends to it, keeping the session id.
- **Activity reporting** via a seeded Pi extension on its own event bus, shelling the generic `agentbox-ctl agent-state pi`.

## Two spec entries that are load-bearing, not cosmetic

- `exclude: ['bin']` — Pi vendors `fd`/`rg` there as **host-native** binaries. Without this a macOS `rg` ships into a Linux box and Pi's grep/find tools are present-but-unexecutable.
- `launchFlags: ['-a']` — Pi's project-trust prompt otherwise blocks an interactive launch into any repo carrying a `.pi/` dir. (This is trust, not tool permission.)

## Two generic fixes the live smoke forced out

Both in the shared pull layer, neither Pi-specific:

1. The default pull ignored `AgentPullSpec.categories`, so an agent whose config is directories-of-items could never pull a box-created item back once the host had that directory. Claude had the same latent gap outside its own module.
2. The pull offered agentbox's **own seeded files** back to the host — accepting one would have installed our extension into the user's real Pi. Now derived from `seeds[].destRel`.

Plus `credential-reconcile.test.ts` was machine-dependent (overrode three agents by name, let the rest read the real `~/.agentbox`).

## e2b was broken; fixed in the base

`--provider e2b` produced a box that reported ready and whose agent died instantly. Root cause was not Pi: E2B's upstream `e2bdev/base` ships **Node 20** while every other provider is Node 24. Pi needs `>= 22.19` (`fs.globSync`), and `npm install -g` only *warns* on an engines mismatch — so the install "succeeded" and the session exited with `SyntaxError: ... does not provide an export named 'globSync'`.

Fixed in the base rather than per-agent (the box's Node is a property of the image; a per-agent workaround would repeat for every future agent, including community ones that can't patch our bake scripts). Two wrong turns are recorded in the script: nodesource-over-apt loses to `/usr/local/bin/node` on PATH, and extracting the tarball on top merges two npm releases into `Class extends value undefined`.

Pi's `postInstall` now also fails loudly on Node < 22 instead of deferring the failure to launch.

## A guard that a two-letter id broke

`no-agent-named-exports.test.ts` matched `[A-Za-z_]*(?:claude|codex|opencode)` — safe only because those names are long. A bare `pi|Pi|PI` alternative matched **60** innocent exports (`ping`, `pickFreePort`, `SPINNER_FRAMES`, `ApiErrorCode`). The pattern now carries a boundary rule for short ids, with both the catches and the rejects pinned.

## Verification

Smoke-tested end to end on **all six providers** — node/pi version, seeded extension, and a real authenticated turn in each: docker, vercel, daytona, hetzner, digitalocean, e2b. All five cloud boxes reported `pi:idle` in `agentbox list`. Teleport checked against the host transcript rather than taken on trust. All boxes destroyed, no orphans.

`pnpm build` / `typecheck` (45/45) / `lint` (24/24) / `test` (44/44) green.

## Known, deliberately left as-is

Pi's `auth.json` is **multi-provider OAuth with rotating refresh tokens**, and the spec declares no `credential.freshness` (my original justification — "API keys don't rotate" — was wrong). Last-writer-wins can therefore drop a whole provider from the backup. `freshness.jsonPath` names a single numeric path and can't express "newest across providers", so the honest fix is a merge rule. Flagged and parked by request, not overlooked.

https://claude.ai/code/session_017bjSm3o3JDVXLLJgvva32P

---

> [!NOTE]
> **Medium Risk**
> Touches agent launch paths (queue, dashboard, control plane, hub validation) and credential/custody flows; many per-agent tail edits can fail silently if a future agent skips them, and Pi’s multi-provider `auth.json` merge behavior is a known gap.
> 
> **Overview**
> Adds **Pi** (`@agentbox/agent-pi`) as a first-class agent alongside Claude, Codex, and OpenCode: registry/config registration, `agentbox pi` (create/attach/start/login), queue `-i` jobs, `agentbox download pi`, dashboard shortcuts **`[p]`**, and runtime staging of `pi-agentbox-extension.js` for in-box activity reporting.
> 
> The integration follows the existing agent-package pattern (CLI spec + docker runtime, no skip-permissions flag; login is interactive `/login` in Pi’s TUI). Several **shared** call sites were generalized so Pi isn’t special-cased forever: control-plane `--agent` validation and box `agent` registration use `AGENT_SYNC_SPECS` / `isAgentKind`, and `inspect` renders extra agents from the sync registry instead of adding another named block.
> 
> Docs, hub OpenAPI/UI, custody paths, and tests (including a stricter `no-agent-named-exports` rule for the two-letter `pi` id) were updated across the monorepo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0d09b316893ef2eb2fb5ea3eb85db10d561b0e7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #356, opened 2026-09-01
feat(hub): serve the agent catalog on GET /api/v1/agents

madarco:
Every client that offers a choice of agent hardcodes the list — the tray's New Box dialog, the web create modal, the create validator. That was already wrong once Pi landed, and it makes an `agentbox agent add` plugin agent unofferable no matter how well the registry knows it.

This serves the registry instead, the way `GET /api/v1/providers` already serves each provider's descriptor.

```
GET /api/v1/agents
{ "agents": [ { "id": "claude", "label": "Claude", "installed": true }, ... ] }
```

### What `installed` means

The host holds the agent's config dir (`spec.staticPaths`) **or** a saved AgentBox login (`~/.agentbox/<id>-credentials.json`). It lets a picker stop offering an agent the user never set up. It is a **hint, not a gate** — an agent installs on demand inside a box, so creating with an uninstalled one still works.

A `state` static path deliberately does *not* count as setup: OpenCode's `~/.local/state/opencode` rides along with the config volume but never enables it (`create.ts`'s `wantOpencode`), so counting it would advertise an agent create wouldn't actually configure. Driven off the registry's own `stagedAs` marker rather than an `id === 'opencode'` branch, so a future agent with a state dir gets the same treatment for free.

### `POST /boxes` accepts what the catalog offers

`parseCreateBox` gained an `allowedAgents` argument; the route passes the live registry ids. Offering an agent the create path then rejects would be a bug. The hardcoded list stays as the default/fallback for the hosted plane, which carries no registry code — so nothing regresses there.

### Shape

The catalog is read in the custom server's scope and handed over the existing `__AGENTBOX_HUB_SYSTEM` seam, because the registry lives behind `@agentbox/sandbox-core` (execa) and a route can't import it in the standalone build. Seam absent → the route serves the built-ins and **omits** `installed`, which clients read as unknown rather than false.

### Verified

- `pnpm build`, `pnpm typecheck`, `pnpm lint` clean; 154 hub tests pass (new `agent-catalog.test.ts`, extended `validate.test.ts`, `openapi-coverage` satisfied).
- Live against a restarted hub: all four agents `installed: true`; with OpenCode's config/data dirs and credential backup moved aside, `opencode` flips to `installed: false` and back on restore. The hidden `example` canary is correctly absent.

### Note on the base

This branches off `feat/custom_agents` (the Pi agent), which is pushed but has no PR of its own, so this PR carries those 5 commits too. Bugbot reviewed that work already at `58b01837`.

The tray-side consumer is a separate change in `madarco/agentbox-tray`.

https://claude.ai/code/session_01TUTPjk1mXxRru4beLcDKun

---

> [!NOTE]
> **Low Risk**
> Read-only discovery endpoint plus aligned box-create validation; hosted-plane behavior unchanged via fallbacks. Main risk is clients misreading absent `installed` as false rather than unknown.
> 
> **Overview**
> Adds **`GET /api/v1/agents`** so clients can discover coding agents from the hub instead of hardcoding names. On a local hub with the custom server, the list comes from the agent registry (including plugin agents from `agentbox agent add`) and includes an optional **`installed`** hint when the host has config dirs or AgentBox-saved credentials — not a create gate. Registry data crosses the existing **`__AGENTBOX_HUB_SYSTEM`** seam from `server.ts` via new **`collectAgentCatalog`**; without the seam (hosted plane), the route returns built-in agents only and omits **`installed`**.
> 
> **`POST /api/v1/boxes`** now validates **`agent`** against the same live registry ids (plus **`none`**) so advertised agents are creatable; **`parseCreateBox`** takes an **`allowedAgents`** override while defaulting to the compiled-in list for the plane path. OpenAPI, API docs, and **`visibleAgentSpecs`** re-export from sandbox-core round out the surface; tests cover catalog **`installed`** semantics (e.g. ignoring **`stagedAs: 'state'`** paths).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0f8b92b0357e06507a7e9efac0d6938f679c3ae. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Pushed `1aaa7d06`: the hub's own web create picker now reads `GET /api/v1/agents` too, so a plugin agent is selectable there and not just over the API.

Verified in the browser against a restarted hub: the dropdown lists Claude / Codex / OpenCode / Pi / Empty, and with OpenCode's config, data dir and credential backup moved aside its option renders as **"OpenCode — not set up on this host"** — which only the endpoint can produce, since the built-in fallback carries no `installed` field. Restored afterwards, endpoint back to all-installed.

Note for review: Bugbot's clean pass covered `f0f8b92b` only — it doesn't re-review later pushes.

Also spotted while here: the `AGENTS` table in `apps/hub/components/icons.tsx:138` is another hardcoded agent list, but it has no importers — dead code. Left alone as unrelated.

PR #361, opened 2026-09-02
refactor(hub): put upload/clone behind the API, and make service agents persistent

madarco:
## What

Three commands that did real work in the CLI now drive `/api/v1` and let the hub execute it.

### 1. `agentbox upload` — was not a thin client, now is

`upload.ts` called `providerForBox()` + `uploadWorkspaceToBox()` directly while `POST /api/v1/boxes/:id/upload` sat unused. It now goes through `HubApiClient.uploadBox`. The hub also starts a paused box, so `ensureBoxRunningVia` is gone from the CLI. UX is unchanged: the `(exclude-list mode)` label, the copied count, and the `N host change(s) were SKIPPED…` warning with its paths all come from the route's `{ mode, copied, conflicts }`.

**What the route's host-backend constraint means for a remote control box — and what I did about it.** The route reads the *host* side of the workspace off the machine it runs on. A control box's copy of the project is not the tree you are standing in, so pushing it would be the worst kind of success. `agentbox upload` therefore **refuses** rather than pretending, via `remoteHubHostFileRefusal` (`control-plane/with-hub.ts`), gated on `HubTarget.onThisMachine` and not `mode` — a `hub expose`-d machine is `remote`-shaped while still running here. The message names `agentbox cp`, which already reaches back to the user's machine. Making `upload` itself work remotely needs a client→hub file upload (custody blobs are the existing primitive) and belongs with the IO-plane move `docs/hub-api-single-path-plan.md` defers.

### 2. `agentbox clone` — audited, and the export moved

It already streamed a create job, but it also called `providerForBox()` and did `exportBoxWorkspace` host-side first. **The export belongs behind the API and is now there.** The burden of proof was on leaving it in the CLI and there was nothing to say: the export reads the *source box*, which the hub reaches on every provider, so the CLI copy was just a second implementation the web UI and the tray could not use. `POST /boxes/:id/clone` already did `prepareClone` + `create`, so the CLI is one call plus the job stream, and the response's `name`/`workspace`/`files` are the authoritative values it reports. Consequence, documented: the clone's workspace dir and its project registration land on the **hub's** machine — which is also where the new box runs.

`--persistent` on the clone was accepted and inert; it now reaches the create. It is tri-state (both flags declared), so absent means **inherit the source box** — a clone of an always-on service box is always-on too — and `false` is only ever an explicit opt-out, never "no opinion" overriding `box.persistent`.

### 3. A service agent defaults to persistent

`resolveCreatePersistent` (`packages/core/src/persistent.ts`, beside `persistentRefusal` so the whole always-on policy is one file) returns `true` for a `caps.surface === 'service'` spec, the explicit flag when passed, and `undefined` otherwise — `undefined` and not `false`, because `false` would override a user's `box.persistent`. Derived from the surface, never an agent id: a new service agent gets the default from its registry row alone.

Both create paths read it:
- CLI — `service-action.ts` passes it to `provider.create`, with the `persistentRefusal` check moved **before** the provider module loads (the order `create` uses) so a capped provider says no instead of failing later.
- Hub — `hub-backend.create` derives it from `findAgentSpec(input.agent)`, so a web-UI or tray create gets the same default, and refuses a capped provider before enqueuing rather than leaving a failed job behind.

`--persistent` / `--no-persistent` are on the service command factory (both, so commander leaves the value `undefined`); the CLI-surface fixture diff is exactly those two flags.

### Fixed along the way: `DockerProvider.create` dropped `req.persistent`

A phase-1 gap. The CLI's own docker paths call `createBox()` directly and pass `persistent` themselves, so nothing had exercised the `provider.create` mapping — the first caller that only goes through it (a service agent) got an expendable box with **no error anywhere**. The first `agentbox openclaw` smoke came back `persistent= None`, which is how this was found.

### OpenAPI

`CreateBox` now documents `repoUrl`, `agentArgs`, `startAgent`, `foreground` and `opts` (with `persistent` spelled out), the `agent` field is described as the open registry-validated id it actually is rather than a stale 5-value enum, and `POST /boxes` gains its `409`. The clone `persistent` is no longer "accepted and inert" — it documents the inherit-from-source rule and the `conflict` refusal — and the response carries the resolved value. The upload route states the host-disk constraint. `openapi-coverage.test.ts` green.

## Smoke tested (local hub, rebuilt standalone + `hub restart`)

**`upload` goes through the API** — the work now runs *in the hub process*, which is what `~/.agentbox/hub.log` shows:

```
$ agentbox upload claw-p
synced 1 file(s) from …/svc-a into claw-p:/workspace (exclude-list mode)

$ tail ~/.agentbox/hub.log
[hub] syncing non-git workspace (file overlay; the box wins on conflict)
[hub] sync: 1 copied, 1 unchanged, 0 kept by the box
```

Conflict UX intact — box version kept, host change reported:

```
synced 0 file(s) … (exclude-list mode)
▲  1 host change(s) were SKIPPED to keep the box's version:
     host-new.txt
$ docker exec agentbox-uptest cat /workspace/host-new.txt
box version
```

**`clone` exports hub-side** (`[hub] exporting …` did not exist before):

```
[hub] exporting uptest:/workspace to /home/vscode/.agentbox/clones/uptest_clone
[hub] exported 3 file(s) (exclude-list mode)
cloned uptest -> uptest_clone
  workspace: /home/vscode/.agentbox/clones/uptest_clone (3 file(s))
$ docker exec agentbox-uptest_clone ls /workspace
agentbox.yaml  foo.txt  host-new.txt
```

**`agentbox openclaw` is persistent** — gateway came up, record and listing agree:

```
└  openclaw on claw-p: http://127.0.0.1:32781
$ agentbox list --global
1  claw-p   running*  …
1  claw-np  running   …
* persistent (always-on): never auto-paused or pruned; restarted after a reboot

state.json:  claw-p persistent= True     claw-np persistent= None
GET /api/v1/boxes:  claw-p … True    claw-np … False
```

**`--provider e2b` refuses** (exit 1, nothing created):

```
■  --persistent is not supported on e2b: E2B sandboxes carry a platform session cap
   (1h on Hobby, 24h on Pro) that the host keepalive can extend within, never remove …
```

**Through the hub create path**, `POST /api/v1/boxes` with no `opts` at all:

```
{"jobId":"7867dbdc0d78f27edc"}
  agent = openclaw
  createOpts.persistent = True          # derived server-side

opts.persistent:false  ->  createOpts.persistent = False
provider e2b           ->  HTTP 409 conflict, refusal message, no job enqueued
```

Tests: `@agentbox/core` 113, `@agentbox/hub` 154, `@agentbox/sandbox-docker` 326, `agentbox` 1515 — all green. `pnpm typecheck` on each touched package, and `@agentbox/web` builds (the docs changes).

## What I could NOT put behind the API, and why

- **`upload` from a machine that is not the hub's** — see §1. Refused with a message that names the reason and the alternative, not silently downgraded.
- **`download` stays direct.** It is IO-plane by the plan's own rule and writes to the *caller's* disk, so a hub could not perform it on the caller's behalf at all.
- **The hub queue worker still cannot BUILD a service-agent box.** `POST /api/v1/boxes {agent:"openclaw"}` validates and enqueues a job carrying the right options (verified above), but the worker dies with `unknown agent kind: openclaw`: `toSyncKind` only accepts `BUILTIN_AGENT_KINDS` (the four TUI agents — this also blocks every `agentbox agent add` plugin agent), and `_run-queued-job.ts`'s session dispatch has no branch for a surface with no tmux session. **Pre-existing** — nothing used this path before, because `agentbox openclaw` creates through `provider.create` — and fixing it is the "service agents through the hub queue" work, not this refactor. It is written down in `docs/plans/service-boxes-backlog.md`, flagged in the OpenAPI description and in `api.mdx` so the spec does not over-promise to the tray.

Docs updated in the same change: `apps/web/content/docs/{api,cli,openclaw}.mdx`, `docs/hub-api-single-path-plan.md` (a follow-up section recording the conversion and the host-file boundary), `docs/plans/service-boxes-backlog.md`.

https://claude.ai/code/session_01UNTyXnCqGgEGSkjEKPP6au

---

> [!NOTE]
> **Medium Risk**
> Touches box create/clone/upload paths and persistence defaults across CLI and hub; wrong tri-state handling or remote upload refusal could surprise users, but changes are aligned with existing hub API and fail-closed refusals.
> 
> **Overview**
> Moves **`agentbox upload`** and **`agentbox clone`** off inline provider/sandbox-core work and onto **`POST /api/v1/boxes/:id/upload`** and **`POST /api/v1/boxes/:id/clone`**, so the hub runs export/push/create and the web UI/tray share one path. **`upload`** adds **`remoteHubHostFileRefusal`** when the owning hub is not on this machine (avoids pushing the control box’s workspace). **`clone`** is one hub call plus the existing job stream; **`--persistent`** is tri-state and inherits the source box when omitted.
> 
> **Service agents** (`caps.surface: 'service'`) now default to **always-on** boxes via shared **`resolveCreatePersistent`** in **`@agentbox/core`**, wired through CLI service create, hub **`POST /boxes`**, and clone **`prepareClone`**, with **`persistentRefusal`** for e2b/vercel before create. Service commands gain **`--persistent` / `--no-persistent`**; **`DockerProvider.create`** now forwards **`req.persistent`**.
> 
> OpenAPI and user docs describe host-disk upload limits, clone persistence inheritance, and **`opts.persistent`** on create.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2265085a2ce9cbdf8083e6399aa43cb6275dfd4a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Both Bugbot findings fixed in 5d3ac8e3 — same branch, PR updated.

**1. Clone lost the foreground lane.** The lane choice now travels with the request that needs it, in a pure `cloneCreateInput` builder (`apps/hub/lib/boxes/clone-create.ts`) beside the existing `controlPlaneCreateRequest` — same reason that one is pure: a Next route handler is not reachable from a unit test (no `@/` alias at test runtime), so the decision lives where it can be asserted.

**2. `--into` resolved against the wrong cwd.** Fixed at the boundary rather than the symptom, as asked. The CLI resolves it against the caller's cwd (`resolveIntoDir` in `commands/clone.ts`) and the API accepts only absolute paths, refusing a relative one by name (`parseCloneBox` in the v1 `validate.ts`). `prepareClone` keeps a defensive guard and now `normalize`s instead of `resolve`s, so no code path can consult the daemon's cwd. The absolute check accepts POSIX **and** Windows forms deliberately — the hub may be a different OS from the client that resolved the path.

Same class, found while auditing: an un-expanded `~` (quoted, so the shell never touched it) used to become `<cwd>/~/x` — absolute, so the new check would have waved it through, and a directory literally named `~` would appear. It is now refused with a message saying why: `~` means "home", and whose home is exactly what a cross-machine path cannot answer.

## Tests (all mutation-checked — reverting the fix fails the test)

```
apps/hub    test/clone-create.test.ts   3 tests
apps/hub    test/validate.test.ts      27 tests  (5 new, parseCloneBox)
apps/cli    test/clone-into-dir.test.ts 6 tests
```

Mutation runs, each with the fix reverted in place:

```
# foreground: true removed
  × cloneCreateInput > enqueues the follow-on create in the FOREGROUND lane
  Tests  1 failed | 2 passed (3)

# absolute-path guard disabled
  × parseCloneBox > REJECTS a relative --into: cwd is client state and does not travel over an API
  Tests  1 failed | 26 passed (27)

# CLI returns the raw string instead of resolving
  × resolveIntoDir > makes a relative path absolute against the caller cwd
  × resolveIntoDir > never returns a relative path — the API refuses one
  Tests  2 failed | 2 passed (4)
```

Full suites green: `@agentbox/hub` 162, `agentbox` 1521. `pnpm typecheck` on both. `@agentbox/web` builds.

## Live smoke (standalone rebuilt + `hub restart` first)

The hub daemon's cwd is `/workspace`, which makes the difference observable. Cloning from a *different* directory with a relative `--into`:

```
$ pwd
/tmp/…/scratchpad/clonehome
$ agentbox clone relsrc -y --name relclone --into ./relclone
cloned relsrc -> relclone
  workspace: /tmp/…/scratchpad/clonehome/relclone (1 file(s))

$ ls /tmp/…/scratchpad/clonehome/relclone      # landed where the user asked
a.txt
$ ls -d /workspace/relclone                    # NOT under the hub daemon's cwd
ls: cannot access '/workspace/relclone': No such file or directory

$ # the clone's create job:
relclone foreground = True | status = done
```

And the API refusing a relative path directly, before it even resolves the box:

```
$ curl -X POST .../api/v1/boxes/nonexistent/clone -d '{"into":"./svc-hz"}'
{"error":{"code":"invalid_request","message":"into must be an absolute path (got \"./svc-hz\") — it names a directory on the hub's machine, which has no notion of your working directory"}}
HTTP 400
```

## The cwd / env / $HOME audit

The rule I wrote down (`docs/hub-api-single-path-plan.md`, in the follow-up section): **a cwd, a `$HOME`, an env var and a shell expansion are all client state. Whatever only the caller knows, the caller resolves — and the API accepts only the unambiguous form and rejects the rest, because the web UI and the tray have no cwd at all to fall back on.** The precedent already in the codebase is box refs: the CLI resolves `1` / a name / an id prefix against local state and sends an id.

Checked against it, everything this PR moved:

| Value | Verdict |
|---|---|
| `clone.into` | **Was broken** — fixed above. |
| `clone` `~` prefix | **Same class** — now refused rather than mangled. |
| `clone.name` / `clone.provider` | Pure strings; `sanitizeMnemonic` and the box record decide, no environment involved. |
| `upload` body | Only `includeNodeModules`. No paths cross the wire at all. |
| `upload`'s host path | `box.workspacePath` off the record, which lives on the hub — not client state. And `upload` is already refused unless the hub is on this machine, so record and caller agree by construction. |
| `uploadWorkspaceToBox` / `exportBoxWorkspace` | Grepped both: no `process.cwd()`, no `homedir()`, no `process.env`. Every host path is an explicit argument. |
| Box ref (`agentbox upload 1`, `clone 1`) | Already correct — resolved client-side against the local project index, sent as an id. |
| `POST /boxes` persistent derivation | Adds only a derived boolean from the registry row. No environment input. |
| `hashProjectPath` / `registerProject` for the clone | Moved from the caller's machine to the hub's, and now computed from the hub's own resolved `workspace` — strictly more correct than before, since that is where the directory actually is. |
| Clone default `~/.agentbox/clones/<name>` | Deliberate and documented: resolves in the **hub user's** home, because that is the machine the clone's box runs on. Same for the project registration. Identical to the old behaviour for a local hub. |

The one pre-existing item I did *not* touch, since it is outside this PR: `opts.envFiles` on `POST /boxes` are host paths interpreted by the queue worker on the hub's machine. Same class, but it predates this change and the worker is its only consumer.

PR #364, opened 2026-09-06
feat: pick a box size at create time, in the web hub and the tray

madarco:
## Why

A box's VM size was reachable only from the CLI (`--size`, or `box.size<P>`), and the legal values lived only as prose inside each provider's `sizeDesc`. So there was no way to start a bigger box from the web hub or the tray — the concrete blocker being a desktop box with enough RAM for something like Godot (8 GB+), which no default provides.

`--size` has **no common grammar** — each backend parses it natively:

| provider | grammar | 8 GB | applies at |
|---|---|---|---|
| docker | *ignored* (`box.memory` / `box.cpus`) | — | — |
| hetzner | server-type slug | `cx33` (4·8), `cx43` (8·16) | create |
| digitalocean | droplet slug | `s-4vcpu-8gb` | create |
| vercel | vCPU count, RAM coupled 2 GB/vCPU | `4` | create |
| remote-docker | `cpu-mem` GB | `4-8` | create |
| daytona | `cpu-mem-disk` GB | `4-8-10` | **bake** |
| e2b | `cpu-mem` GB | `4-8` | **bake** |

That variation is what a UI needs told declaratively — and `ProviderDescriptor` already had the field for it (`sizes`), served all the way to both clients and populated by nobody. The rail existed end to end and was empty.

## What changed

- **Descriptor** — `sizes` populated for the six non-docker built-ins (values read off each backend's own parser and default constant, not invented), plus `sizeHint` and `sizeAppliesAt`. `sizeHint`'s *presence* is what says the list is open to a custom value; vercel omits it because `parseVercelVcpus` throws outside its set, so one field does two jobs with no flag to keep in sync.
- **Hub API** — the new fields ride the existing descriptor pass-through. One new route, `POST /api/v1/providers/{id}/size-check`, answers whether a *specific* size needs a re-bake, via the `sizeIgnoredReason` hook daytona and e2b already implement. Without it a UI would either bake on every size change (2–10 min) or hand the backend a value it throws away. The stale openapi `Provider` schema is refreshed while there.
- **Web + tray** — a Size field in the create form, driven entirely by the descriptor: no switch on provider id, so a community provider gets the picker by declaring `sizes`.
- **Plugins** — additive on an optional field, so `SDK_API_VERSION` stays **4**. SDK `4.0.0` → `4.1.0`; **needs a republish**.

## One decision worth flagging

When a size needs a re-bake **and** the base is also stale, the stale-base prompt is *skipped*. Its "use existing image" answer would boot at the baked size and silently discard the size just picked — the exact failure this change exists to prevent.

`force` on the bake is likewise load-bearing: a size change does not move the build-context fingerprint, so an unforced bake no-ops and the box comes up on the old size.

## Verification

- `pnpm test` + `pnpm typecheck` green. New `apps/cli/test/provider-sizes.test.ts` asserts every declared size key parses under its own provider's parser, pins each "(default)" label to the backend's real fallback constant (mutation-checked), and ties `sizeAppliesAt: 'bake'` to exactly the providers implementing `sizeIgnoredReason`.
- **Live**: created a hetzner box at `cx43` from the web modal — the Hetzner API reported 8 cores / 16 GB. Destroyed after; no orphan servers or firewalls.
- The daytona bake-scoped path shows the note carrying the real baked size (`2-4-8`), read back through `/size-check`.
- Tray builds and its reason-trimmer was checked against the real daytona/e2b strings plus a pass-through case. **Not visually verified** — I could not drive the menu-bar UI headlessly.

Tray side is on `agentbox-tray` `nightly` (`54412bb`).

https://claude.ai/code/session_01T83kN2Zdcx6e1C4KKj6bsa

---

> [!NOTE]
> **Medium Risk**
> Changes box provisioning paths (prepare force + create opts.size) and new API surface; mistakes could wrong-size boxes or trigger unnecessary bakes, but behavior is descriptor-driven and size-check is advisory-only.
> 
> **Overview**
> **Create flows can pick VM size from declarative provider metadata**, instead of relying on CLI `--size` alone. Built-in providers now ship populated `sizes` lists plus `sizeHint` (opens a custom value field when present) and `sizeAppliesAt: 'bake'` for Daytona/E2B. Docker still has no size picker.
> 
> The **hub exposes that metadata and a new advisory `POST /api/v1/providers/{id}/size-check`**, which calls each provider's `sizeIgnoredReason` so bake-scoped backends only trigger a re-bake when the chosen size actually differs. OpenAPI **`Provider` paths drop hard-coded provider enums** in favor of string ids (including community plugins), and the schema documents descriptor fields clients should render from.
> 
> The **web create modal** adds an Advanced **Size** control (preset + Custom when `sizeHint` exists), resets size on provider change, debounces size-check for bake providers, and routes submit through **prepare with `{ size, force: true }`** when needed—skipping the stale-base "use existing" path when size forces a bake so the pick is not silently dropped. Create passes `opts.size` only for create-scoped providers.
> 
> **Guardrails and docs:** `apps/cli/test/provider-sizes.test.ts` ties descriptor keys to each backend's parsers/defaults; provider SDK **4.1.0** and docs describe how plugins declare sizes. Base branch moves out of Advanced in the modal layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daaf2356826bbdad8d28a45d678b3a8ac771af87. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -9,6 +9,18 @@ Entries are generated from the commit history with `/release-notes` and then
 hand-reviewed — they describe what changed for someone using the `agentbox`
 CLI, not the raw commits.
 
+## [Unreleased]
+
+### Added
+
+- **Pick a box size when you create one, in the web hub and the macOS tray.** The
+  choices come from the provider (`cx43` on Hetzner, `4` vCPU on Vercel, `4-8-10`
+  on Daytona — there is no common grammar), with a custom-value field where the
+  backend accepts free-form sizes. On Daytona and E2B, where the size is fixed
+  when the base is baked, choosing a different one rebuilds the base first and the
+  form says so. A community provider gets the same picker by declaring `sizes` on
+  its descriptor.
+
 ## [0.30.0] - 2026-09-05
 
 ### Added
--- a/apps/cli/test/provider-sizes.test.ts
+++ b/apps/cli/test/provider-sizes.test.ts
@@ -1,0 +1,105 @@
+/**
+ * Guard the descriptors' declared `sizes` against the parsers that have to
+ * accept them.
+ *
+ * A size key is a literal `--size` value for one backend, so a typo here is
+ * invisible until someone picks it in the create form and the box fails to
+ * provision. Three providers parse the spec locally and can be asked directly;
+ * hetzner and digitalocean take opaque slugs their API validates, so those get
+ * a shape check only.
+ *
+ * Lives in apps/cli because `@agentbox/config` cannot depend on the provider
+ * packages (they depend on it) — same reason as `provider-descriptors.test.ts`.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { PROVIDERS } from '@agentbox/config';
+import { parseDaytonaSize } from '@agentbox/sandbox-daytona';
+import { parseE2bSize } from '@agentbox/sandbox-e2b';
+import { parseVercelVcpus } from '@agentbox/sandbox-vercel';
+
+function descriptor(name: string) {
+  const d = PROVIDERS.find((p) => p.name === name);
+  if (!d) throw new Error(`no descriptor for ${name}`);
+  return d as (typeof PROVIDERS)[number] & {
+    sizes?: readonly { key: string; label: string }[];
+    sizeHint?: string;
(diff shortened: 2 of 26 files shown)


madarco:
bugbot run

PR #370, opened 2026-09-07
fix(cloud): give a cloud box's agent its workspace, and stop three AppleDouble leaks

madarco:
Blockers 1 and 4 of `docs/plans/service-boxes-plan.md`. Two independent fixes,
one commit each.

## A cloud box's agent never saw its workspace

Measured on a Hetzner OpenClaw box: the user's files were in `/workspace` and
the gateway was pointed at `~/.openclaw/workspace`. The headline "your project
dir is the agent's workspace" behaviour, broken on every VPS provider.

`spec.boxRunEnv` was built and handed to `provision` correctly, then dropped
twice:

- hetzner's `cloudInitBoxEnv` keeps only `AGENTBOX_`-prefixed keys — **rightly**,
  since `req.env` also carries forwarded provider API keys and cloud-init writes
  box.env `0644`;
- and `kickCloudBootstrap` rewrites box.env with `tee` anyway, on every create
  **and** resume, so anything cloud-init wrote is gone a step later.

The bootstrap is the last writer and the shared cloud path, so it is the one
place to fix. The run-env now rides both surfaces it already maintains:

- **`env`** — exported before `agentbox-ctl bootstrap`; the daemon is spawned
  with `env: process.env` and hands each task `{ ...process.env }`. This is what
  actually fixes `openclaw-onboard`.
- **`boxEnvFile`** — because a cloud login shell does **not** inherit the
  daemon's env, so a hand-run `openclaw` would otherwise disagree with the
  service unit. Docker needs only one surface because `docker run -e` serves
  both; a VPS has no equivalent store.

It takes `buildCloudBoxRunEnv`, **not** `buildForwardedEnv` — only what a
registry row declares, never the API keys that would then land in a
world-readable file.

## Three more tars leaking AppleDouble stubs

`tar` on macOS emits `._*` resource-fork stubs without `COPYFILE_DISABLE`, and
they land wherever the archive is unpacked. Three creating call sites still
lacked it: the **no-git cloud seed** (what an OpenClaw workspace usually is — a
real box came up holding `._agentbox.yaml`, `._AGENTS.md`, `._skills`), and both
`--with-env` file packers.

This is the **fifth** time this class has been fixed, because the rule lived only
in the comments of the sites that already had it. A source scan now asserts it:
every tar that CREATES an archive passes `COPYFILE_DISABLE`; extractions are
exempt. It names the offending file:line when it fails.

## Verified on a fresh Hetzner box

```
[...]

PR #371, opened 2026-09-07
feat(openclaw): tell the gateway which box it is running in

madarco:
An openclaw box got no box facts and no setup skill. Its registry row declares no `seeds`, so an agent that can code, run scripts and move files had no idea it was in a sandbox, that `/workspace` is box-local, or that `agentbox-ctl cp` is how a file reaches the user's machine — while claude gets `/etc/claude-code/CLAUDE.md` + the setup skill, and codex gets the same facts folded into `AGENTS.override.md`.

## What it does

One new supervisor task, `openclaw-agentbox-env`, between onboard and render:

```
openclaw-onboard -> openclaw-agentbox-env -> openclaw-render -> openclaw (service)
```

- **Skill** at `/opt/agentbox/skills/agentbox-setup/SKILL.md`, registered via `skills.load.extraDirs` — outside the workspace, lowest precedence, so a user skill of the same name wins and nothing travels with `clone`.
- **Box facts** at `/workspace/.agentbox/AGENTS.md`, derived from the per-provider `/etc/claude-code/CLAUDE.md` and loaded by the bundled `bootstrap-extra-files` hook.

It re-runs on every boot rather than `runOnce`: the prompt file lives in the workspace, so a clone carries the *source* box's facts into the new box, and regenerating overwrites them before the gateway reads them. It runs after onboard (which writes the config file it patches) and before render, so the user's `openclaw:` overlay still has the last word on any shared key.

## Three shapes that were measured, not assumed

Each of these silently produced nothing when it was wrong, against openclaw 2026.9.2:

- The skill must be a **directory** with `SKILL.md`; a flat `agentbox-setup.md` in the same root was not discovered.
- It must be **copied, not symlinked** — openclaw rejects a skill symlink whose real target is outside the source root.
- The prompt file must be a **real file inside the workspace** with one of six canonical basenames; the hook realpath-checks every path, so a symlink out is refused.

## Keeping it out of your repo

`agentbox download` selects with `git ls-files --cached --others --exclude-standard` and deliberately does not apply the exclude list in git mode (dropping a tracked `.claude/` would be a regression), so an untracked `.agentbox/` would have been pulled into the user's project.

The first cut used the box's `.git/info/exclude`. A live box showed that cannot work: `/workspace/.git` in a docker box is a linked-worktree **file**, and git does not read a per-worktree `info/exclude` anyway — the common dir it *does* read is the user's bind-mounted host repo, so writing there would have edited their repo to hide our file. The second commit moves it to the pull layer as a pathspec (`GIT_MODE_EXCLUDE_DIRS`), which covers every provider and both call sites.

## Verification

Unit: 9 tests pin the task (needs-chain, no `runOnce`, directory-not-flat, copy-not-symlink, sentinel + atomic write, the exact config payload, canonical basenames, guarded/`exit 0`) plus one on the pathspec — all mutation-checked. The `sandbox-core` host-path guard was tightened rather than relaxed: an **absolute** `/.agentbox/` must be rooted at `/workspace`, and injecting `/home/ci/.agentbox/…` still fails it.

Live, on a docker box against a git workspace:

- openclaw's own `/context detail` lists `.agentbox/AGENTS.md` among its **injected workspace files** (3,001 chars, alongside the workspace's own 7,926-char `AGENTS.md`) and `agentbox-setup` among its 38 skills — the file reaches the model, not just the disk.
- The in-box listing reports `.agentbox/AGENTS.md` without the pathspec and nothing with it, and a real `agentbox download` left the repo clean.

Not covered: the cloud providers were not re-run for this. The task is provider-agnostic (one ctl task, riding the `agents.list` RPC), but the prompt half depends on `OPENCLAW_WORKSPACE_DIR` reaching the box, which lands in #370.

---

> [!NOTE]
> **Medium Risk**
> Boot-time OpenClaw config patching and workspace pull exclusions affect every OpenClaw box; mistakes could clobber user skill/bootstrap paths or leak box-local prompts into repos, though merges are defensive and user overlay still runs last.
> 
> **Overview**
> OpenClaw boxes now get the same **sandbox context** Claude/Codex already receive: a per-boot supervisor task **`openclaw-agentbox-env`** runs after onboard and before config render.
> 
> It **copies** the baked setup guide into `/opt/agentbox/skills/agentbox-setup/SKILL.md` and registers that dir via a **union merge** into `skills.load.extraDirs` (so user dirs are not wiped on the second boot). It **regenerates** `/workspace/.agentbox/AGENTS.md` from the provider box facts and adds that path to the `bootstrap-extra-files` hook the same way. Patching is gated on `openclaw config validate` and uses `openclaw config get` plus an in-box Node merge program because `config patch` replaces arrays wholesale and the config file may be JSON5.
> 
> **Sync/pull** now drops `.agentbox` everywhere it would leak generated box facts into the user's repo: exclude-list mode, clone, and git-mode `download` via `GIT_MODE_EXCLUDE_DIRS` pathspecs. Docs describe the two injected artifacts and the task ordering; tests cover the merge program, task script shape, and workspace listing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3706bec2019749abbd860e923067e93694884ca. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/content/docs/openclaw.mdx
+++ b/apps/web/content/docs/openclaw.mdx
@@ -158,9 +158,9 @@ Two things, both AgentBox's rather than yours, refreshed on every boot:
   description Claude and Codex get, so the gateway's agent knows it is in a
   sandbox, that `/workspace` is box-local, and how to reach your machine. It has
   to live inside the workspace because OpenClaw reads a system prompt from
-  nowhere else, so AgentBox adds `.agentbox/` to the box's own
-  `.git/info/exclude` — it will not follow an `agentbox download` into your repo,
-  and your own `AGENTS.md` is never touched.
+  nowhere else, but `agentbox download` and `agentbox clone` both skip
+  `.agentbox/` — it will not follow a pull into your repo, and your own
+  `AGENTS.md` is never touched.
 
 Both are regenerated each time the supervisor starts, which is what keeps a
 cloned box from inheriting the facts of the box it was cloned from.
--- a/apps/web/content/docs/openclaw.mdx
+++ b/apps/web/content/docs/openclaw.mdx
@@ -147,6 +147,24 @@ The unit names (`openclaw`, `openclaw-onboard`, `openclaw-render`) are public su
 
 Only one unit may `expose:`. If your `agentbox.yaml` already publishes the box's web port, OpenClaw's `expose:` is dropped with a warning rather than silently taking it over.
 
+## What AgentBox adds to the box
+
+Two things, both AgentBox's rather than yours, refreshed on every boot:
+
+- **A skill** at `/opt/agentbox/skills/` — outside your workspace, registered
+  through OpenClaw's `skills.load.extraDirs` at its lowest precedence, so a skill
+  of your own with the same name wins. It never travels with `agentbox clone`.
+- **The box facts** at `/workspace/.agentbox/AGENTS.md` — the same sandbox
+  description Claude and Codex get, so the gateway's agent knows it is in a
+  sandbox, that `/workspace` is box-local, and how to reach your machine. It has
+  to live inside the workspace because OpenClaw reads a system prompt from
+  nowhere else, so AgentBox adds `.agentbox/` to the box's own
+  `.git/info/exclude` — it will not follow an `agentbox download` into your repo,
+  and your own `AGENTS.md` is never touched.
+
+Both are regenerated each time the supervisor starts, which is what keeps a
+cloned box from inheriting the facts of the box it was cloned from.
+
 ## State, sync and clones
 
 | Path in the box                   | What it is                      | Travels?                         |
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -381,6 +381,17 @@ declares instead of the TUI machinery:
 - **`service.urlFields`** — values `<agent> url` prints beside the URL, read out
   of the daemon's own config file (a Control UI's gateway token). Data, because
   the only per-agent parts are which file and which dotted key.
+- **`service.tasks`** is also where AgentBox asserts config it OWNS. openclaw's
+  `openclaw-agentbox-env` task installs the AgentBox skill under
+  `/opt/agentbox/skills` (registered via `skills.load.extraDirs`, outside the
+  workspace so a clone never carries it) and writes the box facts to
+  `/workspace/.agentbox/AGENTS.md` (registered via the `bootstrap-extra-files`
+  hook, which refuses any path outside the workspace and any name that is not
+  one of openclaw's six canonical bootstrap basenames). It runs on EVERY boot,
+  because the workspace file travels with `agentbox clone` and would otherwise
+  describe the box it came from. `configRender` is not the seam for this: it
+  reads only the user's `agentbox.yaml` block, and the render runs after this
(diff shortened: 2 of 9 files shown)


madarco:
Fixed in 331caf2b — good catch, this was real.

`openclaw config patch` replaces arrays wholesale and `openclaw-render` only re-sends *changed* overlay keys, so a stable user value survived boot 1 and was reverted on boot 2. The task now computes its patch in the box from the live config and unions our entry in, so AgentBox asserts membership rather than a value: other entries, neighbouring keys and the rest of the hook entry are untouched, and an explicit `enabled: false` is preserved.

Verified live: `openclaw config set skills.load.extraDirs '["/home/vscode/my-skills","/opt/agentbox/skills"]'`, stop, start — the task re-runs and both values are still there.

bugbot run

madarco:
bugbot run

madarco:
Fixed in e3cffed7 — right again, and it reopened the first finding through another door.

Confirmed against a live box: openclaw reads a config with a `//` comment in it fine, and `JSON.parse` throws on that same file. So the merge would have seen `{}` for any commented config and clobbered both arrays.

The merge now asks openclaw for the current values (`openclaw config get`) instead of parsing the file — its own reader, so JSON5, defaults, profiles and env overrides all come out right. It is gated on `openclaw config validate` because `config get` prints nothing and exits 1 *both* for an unset value and for one it cannot read: without the gate a broken config looks exactly like a fresh box and gets overwritten.

Verified live: a config `JSON.parse` throws on, holding a user's own `extraDirs` entry, survives a stop/start with both values intact.

bugbot run

madarco:
bugbot run

madarco:
Fixed in e4beb06e.

One correction on the premise: `entries['bootstrap-extra-files']` *does* parse on openclaw 2026.9.2 — I checked it in a box before writing it, and it returns the entry. So there was no live bug here.

The failure mode you point at is real though, and the `config validate` gate does not cover it: it says the config is good, not that an exotic path syntax parses, and a failed get is indistinguishable from an unset one. So the reliance is gone rather than defended — the task now reads the whole `hooks.internal.entries` object by plain dot path and picks our key out in the program. Naming only our key in the patch also keeps neighbouring entries out of it.

Verified live: a box with a user entry in *both* arrays, `session-memory` sitting alongside our hook, and a config `JSON.parse` throws on — all intact across a stop/start.

madarco:
bugbot run

madarco:
d3706bec — one of the two was mine, one is stale.

**Idempotency test uses stale argument shape** — correct, and a real hole. The fixture passed the inner hook object where the program now expects the whole `entries` map, so the lookup missed and the case quietly re-tested the *unset* path, where a duplicate `paths` entry can never appear. It would have passed against a merge that appends blindly; it now fails against one (checked by mutation). The edit that should have updated it matched nothing after prettier joined the line.

**Hook get path can clobber config** — this one is stale: it describes `entries['bootstrap-extra-files']`, which e4beb06e had already replaced with the plain dot path `hooks.internal.entries` (line 200), precisely to remove that reliance. Nothing in the current tree reads a bracketed path.

madarco:
bugbot run

PR #372, opened 2026-09-07
fix(sync): make the workspace selection and the staging agree

madarco:
`agentbox download` failed outright and transferred **nothing** on a git workspace whose repo does not gitignore `node_modules` — 616 `link_stat … No such file or directory` errors, then `rsync error … (code 23)`. Found while answering a question about how download works.

## Why

The selection and the staging were derived independently, and any disagreement was fatal:

- the **selection** is `git ls-files --cached --others --exclude-standard`, run in the box;
- docker's **staging** mirrors `/workspace` with `rsync -a --delete --exclude=node_modules`, and has never known about `workspaceExcludes()`;
- `rsyncPullToHost` dry-runs first and **threw** on any non-zero exit, so one listed-but-absent path killed the download during the *preview* pass.

The same shape had two more live triggers, both verified:

- `git ls-files --cached` prints an index entry whose worktree file was **deleted**. On the cloud path tar fails that member under `set -e`, so a box that deleted a tracked file broke `download` **and** `clone` on every cloud provider — the same bug on the other transport.
- Docker's apply pass re-lists the live box while reusing the preview's scratch dir, and `--no-refresh` widens that to anything created since the last staging.

## What changed

1. **The selection may only name paths that exist.** The git branch pipes through an existence test (`xargs -0 sh -c`, NUL-safe; no `/dev/fd`, which a Vercel box lacks; 2.65s vs 8.6s for a bash read-loop over 138k paths). `-f`/`-L` mirrors the find branch and also drops gitlink **directories**, which tar recursed and rsync did not. `set -o pipefail` is load-bearing — without it a failing `git ls-files` exits 0 and the short list reads as a smaller workspace.
2. **`node_modules` leaves the git-mode selection unless asked**, making `--include-node-modules` one switch in both modes. It needs **three** pathspecs, not one: rsync's `--exclude` matches a basename at any depth while a pathspec is root-anchored, and the root-anchored form alone still lists **3310** nested paths in this repo (git 2.39).
3. **A residual mismatch warns instead of aborting.** macOS ships rsync 2.6.9, which has no `--ignore-missing-args`, so the list is reconciled with the staged copy. A pull is additive, so a raced file is reported and skipped; an *empty* intersection is not a race but stale staging, and still fails, naming the scratch dir.

Docker's mirror and the selection now read the same flag through one `exportMirrorExcludes`, documented with the invariant they must satisfy — **the mirror is a superset of the selection** — with a drift test that makes it mechanical rather than remembered.

Also, per the decision in the thread: agent-generated workspace files (openclaw's `AGENTS.md`, `SOUL.md`, `IDENTITY.md`, `USER.md`, which it writes because its workspace *is* the project) are neither silently dropped nor silently copied. `download` **asks** when they would be new to the project, `--include-agent-files` answers yes up front, `-y` skips and says so, and a file you already have is never questioned. `clone` drops them, matching its existing fresh-identity contract.

## Verification

Unit: a behavioural test builds a real git repo carrying every hazard (deleted-tracked file, `node_modules` at two depths, a name with a space, a symlink, a dangling symlink) and **executes** the generated script — it fails on today's code. Plus real-rsync tests for the skip/fail-loud/skipPaths paths, the 9-vs-11-char itemized parser, and the mirror drift guard. Every guard mutation-checked.

Live, docker (git workspace): the dry run that exited 23 now succeeds; 630 git-listed paths become 7 changes with no `node_modules`; the deleted tracked file is absent; a wiped scratch dir + `--no-refresh` fails with an actionable message instead of exit 23; a file removed after staging is named rather than fatal.

Live, docker (**gitless** `express-server`): exclude-list mode drops `node_modules`, `media/`, `state.sqlite` and `.claude`, and the agent-file prompt works there too.

Live, **e2b**: `git ls-files --cached` lists `foo.txt`, it is not on disk, and the new selection returns 0 for it — the cloud tar abort fixed at the source.

One note: the two new subprocess-spawning suites carry an explicit 30s timeout. They shell out to real `rsync`/`git`, take ~700ms idle but 1.7s under load, and against vitest's 5s default they failed twice in a full `build && test` run while never failing alone.

Not addressed here: docker's apply pass re-lists live while the cloud path replays its sidecar — giving docker the same sidecar would remove that divergence entirely.

---

> [!NOTE]
> **Medium Risk**
> Changes core workspace pull/sync paths used by download and clone on every provider; behavior is safer (fewer total failures) but alters what gets copied by default for node_modules and openclaw workspace files.
> 
> **Overview**
> Fixes **`agentbox download`** failing with rsync exit 23 when the file **selection** and docker **staging mirror** disagreed — most visibly git repos that do not gitignore **`node_modules`**, where git listed paths the mirror had already excluded.
> 
> **Selection and mirror now share one rule for `node_modules`:** git-mode uses depth-aware pathspecs (not root-only), docker’s export mirror reads the same **`includeNodeModules`** flag via **`exportMirrorExcludes`**, and **`--include-node-modules`** applies in both git and exclude-list modes. The in-box list script only emits paths that **exist** (filters deleted index entries, adds **`pipefail`**), avoiding fatal tar/rsync on “listed but absent” files.
> 
> **Stage 2 (`rsyncPullToHost`)** reconciles **`--files-from`** with the scratch dir: missing paths are **skipped and reported** instead of aborting the whole pull; an empty staged set still fails with a clear stale-**`--no-refresh`** message. **`skipPaths`** supports declining specific files after preview.
> 
> **CLI UX:** **`--include-agent-files`** and an interactive prompt (or **`-y`** skip) for **new** agent-generated workspace files (**`AGENTS.md`**, **`SOUL.md`**, etc., declared per agent in the registry); **`clone`** continues to drop those artifacts for fresh identity. Docs and integration tests cover mirror drift, real git selection, and real rsync behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a0927be7caa6f6d5a913039baefa643a321e0b2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #373, opened 2026-09-07
feat(sync): back a bot up into its own project

madarco:
Phase 1 of [`docs/plans/bot-clone-spawn-plan.md`](docs/plans/bot-clone-spawn-plan.md).

## What it adds

`agentbox download --backup [box]` captures a box as files that can recreate it on any provider — the thing a checkpoint cannot do, because a checkpoint is a provider-native snapshot and does not travel from E2B to Hetzner.

```
<project>/.agentbox/bots/ada/
  latest -> 2026-09-07T15-09-16Z
  2026-09-07T15-09-16Z/
    manifest.json
    workspace/        # the same pull a plain download runs, pointed elsewhere
    state/            # the agent's state dir, IDENTITY INCLUDED
```

Timestamped, newest three kept (`--keep`). `--name` sets the bot name, `--agent` picks whose state to take.

The state half deliberately does **not** reuse `staticPaths[].exclude`: that is push-direction hygiene and for openclaw it is exactly the identity — a capture that dropped `openclaw.json` would restore a different bot. A new data-only `stateBackup` spec field says what a capture must leave behind instead (the uid-keyed lock dir, whose name differs per provider). Live databases go through SQLite's online-backup API.

## The bug it forced first

Putting a backup in the project is only safe if the project's `.agentbox/` never reaches a box. It did. **Reproduced live** on a box created with the filter reverted:

```
$ docker exec agentbox-leakbox cat /workspace/.agentbox/bots/ada/*/state/openclaw.json
SECRET-GATEWAY-TOKEN
```

The pull already dropped `.agentbox`; the seed never did. Now every host→box workspace path drops it, derived from the one list the pull reads: both no-git tar seeds, all three `git ls-files --others` carry-over lists (docker, cloud, custody), and the APFS snapshot's prune set. The cloud tar seed also gained the `COPYFILE_DISABLE` it was the last tar in its file to be missing (blocker 4 in the service-boxes plan).

## Measured, not assumed

- **Neither tar anchors an exclude.** bsdtar 3.5.3: `--exclude=./.agentbox` still drops `sub/.agentbox`. The seed matches at any depth — which is also right, since a backup taken from a monorepo sub-project writes `packages/foo/.agentbox/`. The git-mode pull pathspec stays root-anchored, so the seed refuses slightly more than the pull returns.
- **`git check-ignore .agentbox` exits 1** against a `.agentbox/` rule: a trailing-slash pattern matches directories only and the path does not exist yet. The probe is a path inside the dir, so it works whichever way the user spelled their rule.
- **The write-ahead log matters.** A live gateway had a 1.5 MB `openclaw.sqlite` beside a **1.94 MB** `-wal`. The capture is 3.27 MB, `pragma integrity_check` ok, 326 tables. A byte copy would have silently dropped 1.9 MB of committed writes.
- **`find` discovered a database nobody would have listed by hand**: `agents/main/agent/openclaw-agent.sqlite`.
- rsync creates only the last component of its destination, so the bundle dirs are made before the pull, dry-run included.

## Verified live

| Check | Result |
|---|---|
| git project, untracked un-ignored backup | not in the box; **leaks with the fix reverted** |
| gitless project (tar exclude is the only guard) | not in the box |
| e2b create from the same project | not in the box |
| openclaw backup | token matches, `tmp/` dropped, 2 databases captured, integrity ok |
| workspace half, retention, `latest`, gitignore | all correct; `git status` shows only the `.gitignore` edit |
| e2b backup | workspace captured, manifest records the provider |

The **state** half is docker-only for now, deliberately: openclaw is refused on e2b/vercel (session caps) and on hetzner/DO still hits blocker 1 in the service-boxes plan. Not a claim of cloud parity.

4815 tests pass; every new guard is mutation-checked. `download` remains the one CLI command not behind `/api/v1`, so the orchestration lives in `sandbox-core` for a hub route to reuse.

---

> [!NOTE]
> **Medium Risk**
> Touches workspace sync/seeding and writes gateway identity to disk; the change fixes a prior leak and adds guards/tests, but backup bundles remain highly sensitive secrets.
> 
> **Overview**
> Adds **`agentbox download --backup`**, which runs the same workspace pull as a normal download but writes under `<project>/.agentbox/bots/<bot>/<timestamp>/` (workspace + optional agent **state**, including gateway identity). Shared logic in **`sandbox-core`** (`bot-backup`) handles manifests, retention (`--keep`, default 3), a `latest` symlink, `.gitignore` for `.agentbox/`, and state capture via **`stateBackup`** on agent specs (OpenClaw keeps identity files push normally excludes; live DBs use SQLite’s online-backup API). The CLI module **`_backup.ts`** wires flags (`--name`, `--agent`) and post-pull orchestration; docker **`pullToHost`** gains a **`destDir`** override so cloud and docker share one download path.
> 
> **Fixes a serious seed leak:** host `.agentbox/` (including backup tokens) was copied into new boxes via no-git tars and git untracked carry-over. Host→box paths now drop `.agentbox` through **`seedExcludeTarArgs`** / **`dropHostOnlyPaths`** (docker, cloud, custody, APFS snapshot prune); cloud no-git seed also sets **`COPYFILE_DISABLE`**.
> 
> Docs and tests cover the invariant, backup behavior, and CLI options.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c099834000b86620c393efb618cd3a18c621df2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
## Now verified on Hetzner, state half included

Testing the backup on a VPS surfaced a prerequisite, so this branch also carries the fix for **blocker 1 in `docs/plans/service-boxes-plan.md`**.

### The bug

`boxRunEnv` reached a box's processes differently on every provider, and on the two VPS ones it did not arrive at all. Docker sets it on `docker run`; Vercel and E2B hand it to the SDK as a sandbox-wide env. A VPS has no such primitive, so Hetzner and DigitalOcean carry it by writing `/etc/agentbox/box.env` — which both filter to `AGENTBOX_*`:

```ts
if (v !== undefined && k.startsWith('AGENTBOX_') && !CLOUD_INIT_BOX_ENV_EXCLUDE.has(k))
```

`OPENCLAW_WORKSPACE_DIR` was dropped there, so `openclaw onboard` ran against `~/.openclaw/workspace` and "your project dir is the agent's workspace" was quietly false. Nothing failed loudly: the gateway still bound its port and passed its health check, so every smoke called the box ready.

This mattered for the backup work directly — a bundle captured on Hetzner would have held a bot that was not working on the project at all.

### The fix, and why not the obvious one

That filter is not a bug: `box.env` is world-readable and the prefix rule is what keeps the relay and bridge tokens out of it. So the fix went the other way. `buildAgentDescriptors` folds `boxRunEnv` into the service **and each of its tasks** — onboard is a task, and it is the one that reads the variable, so a service-only fix would have left the bug in place. The values ride `agents.list` alongside the units that need them: one rule for all five providers, no provider implements anything, non-secret paths stay out of a world-readable file, and **no re-bake**.

A unit's own `env` still wins, since `boxRunEnv` is the agent-wide default.

### Live proof

```
$ openclaw config get agents.defaults.workspace
/workspace

$ cat /etc/agentbox/box.env          # still AGENTBOX_* only — it did NOT come this way
AGENTBOX_BOX_ID=b2c546f01
AGENTBOX_BOX_NAME=hzbot
AGENTBOX_BOX_KIND=cloud
AGENTBOX_WEB_PROXY_PORT=8080
AGENTBOX_BOX_HOST=hzbot.localhost

$ ls ~/.openclaw/workspace
No such file or directory
```

### Backup on Hetzner

| Check | Result |
|---|---|
| identity captured | backup token `2fbdc218e0683fb5…` == the live gateway's |
| uid-keyed `tmp/` | dropped |
| database | 3.28 MB captured from a 1.5 MB main + 1.94 MB WAL; `integrity_check` ok, 326 tables |
| manifest | `provider: hetzner`, `state: true` |
| second box from the same project | no `.agentbox`, and the token appears nowhere in `/workspace` |

Both VPSs destroyed; the Hetzner API shows no orphaned servers or firewalls.

**Still open, deliberately:** a shell you open yourself does not get these keys, so a hand-run `openclaw` on a VPS box disagrees with the service. Widening `box.env` is the fix and needs a rule for which keys may enter a 0644 file — noted in the plan rather than done here.

PR #374, opened 2026-09-07
feat(sync): bring a backed-up bot back into a new box

madarco:
Phase 2 of [`docs/plans/bot-clone-spawn-plan.md`](https://github.com/madarco/agentbox/blob/agentbox/bot-restore/docs/plans/bot-clone-spawn-plan.md). Stacked on #373 (Phase 1, `download --backup`) — review or merge that first.

`--restore <bot>` recreates a bot from a backup bundle:

```bash
agentbox openclaw --restore ada                    # identity included
agentbox openclaw --restore ada --provider hetzner # on another provider
agentbox create   --restore ada                    # the workspace half only
```

## The PoC's answer was a bug, not a yes

The plan opened with a live gate: does pushing `state/` into a fresh box and restarting the gateway actually revive the bot? It does — **but only after the target's stale write-ahead logs are removed.**

A fresh openclaw box has already onboarded, so its `state/` holds a live `openclaw.sqlite` with its own `-wal`/`-shm`. Copying the bundle's main file over it leaves that WAL describing a database it no longer matches, and the gateway then refuses to start:

```
SQLite integrity_check failed for /home/vscode/.openclaw/state/openclaw.sqlite:
  row 1 missing from index idx_gateway_boot_lifecycle_started
```

— on a backoff loop, forever. The naive "rsync the state dir in" implementation ships a bot that never boots. That deletion is now the first thing `restoreAgentState` does, scoped to the databases the bundle actually replaces (a sidecar of one we are *not* overwriting may hold the only copy of committed rows).

## No create-path surgery was needed

The plan expected to write the state in before onboarding. `run_once: 'marker'` makes that both impossible and unnecessary: the marker is keyed by the task's command hash and lives on the box **rootfs**, not in the agent's config volume, so onboard runs once on every fresh box whatever that volume holds. Letting it run and then replacing what it wrote means the marker is already down and onboard never touches the restored identity again — on that boot or any later one, verified across a full `stop`/`start`.

So: create, stop the unit, push, restart. One order, every provider, no provider code touched.

## What else the implementation found

- **The state push must not go through `agentPushExcludes`** — it adds `LIVE_DATABASE_EXCLUDES` unconditionally, which matches `*.sqlite*`, precisely what a restore exists to put back. Pinned by a test.
- **The workspace half needs no plumbing.** The restored tree is the box's project, exactly as `clone`'s export dir is. `projectRoot` is that directory verbatim rather than `findProjectRoot`'s answer: the restore dir lives under the *original* project's `.agentbox/`, so walking up would have seeded the box from the template project instead.
- **A Phase 1 bug this exposed:** `listBackups` counted a restore's live `workspace/` as a backup, and `workspace` sorts after every stamp — so it would have become the newest "backup" the prune must keep.

## Verification (live)

| | |
|---|---|
| docker | same gateway token, the same automation row (**same id**) in the restored database, `openclaw config validate` clean, surviving a full box `stop`/`start` |
| hetzner | the acceptance test: a bundle captured on **docker**, restored onto a **VPS** — same token, same automation, config valid, workspace intact |
| `create --restore` | workspace half only, naming the agent whose identity it did not bring |

Both test boxes destroyed; the Hetzner API shows no orphan servers or firewalls. 4,832 tests pass, typecheck clean across 47 tasks, docs site compiles. Both new guards are mutation-checked.

## Deliberately still open

Recorded in the new [`bot-clone-spawn-backlog.md`](https://github.com/madarco/agentbox/blob/agentbox/bot-restore/docs/plans/bot-clone-spawn-backlog.md):

- **The source-box guard reads `manifest.boxId`**, so a second restore of the same bundle into a different `--into` directory is not refused. Restoring twice into the *same* directory is.
- **"Does OpenClaw refuse a duplicate identity?" is not answerable locally** — two gateways on one host with no channel configured never collide, and the collision that matters is at the channel provider. The hard error with `--force` is the safe default, chosen rather than measured.
- **`--restore` does not work against a remote hub**: the bundle is a local directory and the create API carries a `projectId`, never a path.
- **`download` is still the one CLI-inline command**, and `--restore` joins `--backup` there. Moving the whole command behind `/api/v1` — with a state-push route the restore can use — is one change, not three.

---

> [!NOTE]
> **High Risk**
> Restores gateway identity and SQLite state into live boxes; mistakes could cause duplicate identities or broken gateways, though guards (`assertSourceBoxNotRunning`, WAL clearing) mitigate the main failure modes.
> 
> **Overview**
> Adds **`--restore <bot>`** (with `--stamp`, `--into`, `--force`) so backups under `<project>/.agentbox/bots/` can seed a **new** box. **`agentbox create --restore`** copies the workspace into a live sibling dir (default `bots/<bot>/workspace`) and warns when the bundle also has agent state. **`agentbox <agent> --restore`** does the same plus, after the service is healthy, stops the unit, pushes captured state via **`restoreAgentState`**, and restarts—so gateway identity can return on another provider.
> 
> New CLI module **`_restore.ts`** centralizes bundle resolution, workspace staging, refusals (no positional box ref with restore, non-empty dest, source box still running), and **`stopUnit`** moved to **`service-action.ts`** for reuse.
> 
> In **`sandbox-core`**, **`resolveBotBundle`**, manifest checks, and **`restoreAgentState`** clear stale **`-wal`/`-shm`** before pushing SQLite files (avoids integrity-check boot loops) and bypass push excludes that would drop databases. **`listBackups`** now only counts stamp-shaped dirs so a restore’s live **`workspace/`** is not treated as a backup for prune.
> 
> Docs, changelog, plan/backlog updates, and unit tests cover guards and restore behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14c3545da02f8bd91bfe3a5bb091d2db85cd8017. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -25,6 +25,11 @@ CLI, not the raw commits.
   `<project>/.agentbox/bots/<bot>/<timestamp>/`, so a bot can be recreated later or
   on a different provider. The newest three are kept (`--keep`), and live databases
   are captured through SQLite's online-backup API, not copied byte for byte.
+- **`--restore <bot>` puts a backed-up bot back into a new box.** On a
+  service-agent command (`agentbox openclaw --restore ada`) it restores the state
+  dir too, so the bot keeps its identity, pairings and history — on a different
+  provider if you pass one. `agentbox create --restore` takes the workspace only.
+  Refused while the backed-up box is still running, unless `--force`.
 
 ### Fixed
 
--- a/apps/cli/src/agents/command/service-action.ts
+++ b/apps/cli/src/agents/command/service-action.ts
@@ -341,7 +341,27 @@ export async function runServiceAgent(
         );
       }
       await assertSourceBoxNotRunning(restored.bundle, opts.force);
+      // A restore always makes a NEW box. The bare command is create-or-resume,
+      // so without this a second `--restore` into the same directory would push
+      // a bundle's identity over a running bot's — the one thing this must never
+      // do by accident. Checked BEFORE the copy, or `--force` would overwrite
+      // that box's workspace on its way to being refused.
+      const already = await findExistingBox(undefined, restored.workspaceDir, spec.id);
+      if (already) {
+        throw new Error(
+          `box ${already.name} already runs on ${restored.workspaceDir} — ` +
+            `pass --into <dir> to restore alongside it`,
+        );
+      }
+      // Staged BEFORE the config load below, which reads the workspace: on a
+      // first restore the destination does not exist yet, and a config loaded
+      // from a missing directory silently falls back to the defaults.
+      const staged = await stageRestoreWorkspace(restored, opts.force);
       opts.workspace = restored.workspaceDir;
+      log.info(
+        `restoring ${restored.bundle.bot} @ ${restored.bundle.stamp}: ` +
+          `${String(staged.files)} entr(ies) -> ${restored.workspaceDir}`,
+      );
     }
     const project = restored
       ? { root: restored.workspaceDir }
@@ -353,22 +373,6 @@ export async function runServiceAgent(
 
     intro(`agentbox ${spec.id}`);
     const existing = await findExistingBox(boxRef, project.root, spec.id);
-    // A restore always makes a NEW box. The bare command is create-or-resume, so
-    // without this a second `--restore` would push a bundle's identity over a
-    // running bot's — the one thing this feature must never do by accident.
-    if (restored && existing) {
-      throw new Error(
-        `box ${existing.name} already exists for this restore dir; ` +
-          `pass --into <dir> or -n <name> to restore alongside it`,
-      );
-    }
-    if (restored) {
-      const staged = await stageRestoreWorkspace(restored, opts.force);
-      log.info(
-        `restoring ${restored.bundle.bot} @ ${restored.bundle.stamp}: ` +
-          `${String(staged.files)} entr(ies) -> ${restored.workspaceDir}`,
-      );
-    }
     let box: BoxRecord;
 
     if (existing) {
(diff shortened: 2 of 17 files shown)


madarco:
All three Bugbot findings were real. Fixed in `14c3545d`, and each is now a pure function so it can be asserted without provisioning a box.

**1. Workspace-only restore fails after create (High).** Correct. `restoreStateIntoBox` threw on a bundle with no `state/`, *after* the box was already up — so the command failed and left a working box behind it. `restoreScope(bundle)` now says which halves a bundle carries and the service path skips the state half with a warning, matching what `create --restore` already did.

**2. Restore can overwrite an existing box (High).** Correct, and worse than the pre-copy guard suggested: with a positional ref the later `findExistingBox(boxRef, …)` could resolve a box anywhere, resume it, and then the restore would stop it and write the backup's identity over it. There is no safe reading of `agentbox <agent> foo --restore bar`, so it is now refused before anything resolves — and the existing-box refusal is re-checked after the lookup, as a belt to the pre-copy braces.

**3. Failed WAL cleanup still copies databases (Medium).** Correct. `rm -f` is silent about a file that was not there, so a non-zero exit meant the removal genuinely did not happen — and pushing anyway lands the databases on the leftover logs, which is exactly the restart loop this step exists to prevent. It fails there now: stopping leaves a box with its own working identity, continuing leaves one that never boots.

## Re-verified live after the refactor

| | |
|---|---|
| `openclaw bea --restore bea` | refused, and `bea` was **not** resumed or touched |
| `--restore bea` while `bea` runs | refused by the source-box guard |
| `--restore bea` after destroying it | token `c2d7bb08…` matches, automation row back, config valid |
| `--restore cleo` (workspace-only bundle) | box came up, warned that the identity is fresh, exit 0 |

4,839 tests pass, typecheck clean across 47 tasks, and all three new guards are mutation-checked. Boxes destroyed, no volumes or cloud resources left.

madarco:
bugbot run

madarco:
Merged into `nightly` by fast-forward (`98fc1848..14c3545d`), which is how features land here — the branch tip *is* nightly now, so there is nothing to squash. Closing rather than merging because this PR targeted `agentbox/bot-backup` (#373), which went in with it.

Phases 1 and 2 of the bot backup/restore plan are now on nightly: `download --backup` and `--restore`, both live-verified on docker and hetzner.

PR #375, opened 2026-09-07
feat(clone): spawn a second bot, with its own name and its own secrets

madarco:
Phases 3 and 4 of `docs/plans/bot-clone-spawn-plan.md`. Backup (#373) made a bot capturable and restore brought it back; this makes a *second* bot.

Before this, `agentbox clone` of an openclaw box gave you an agentless directory whose `SOUL.md` still named the source bot, and the only way secrets reached it was a project-wide `carry:` entry both boxes shared — so the copy answered as the original.

## What an agent now declares

```ts
clone: {
  drop: ['AGENTS.md', 'USER.md'],          // it regenerates these
  render: ['SOUL.md', 'IDENTITY.md'],      // the user's words; rewrite the name
  perBoxCarry: [{ src: '~/.agentbox/openclaw/{{AGENTBOX_BOX_NAME}}.env', dest: '~/.openclaw/.env', mode: 0o600, optional: true }],
}
```

Data, not a hook — the hub runs clones, and a package agent reaches it as JSON. `workspaceArtifacts` is replaced by it (unreleased, so no alias); the `download` prompt reads the union, which is the same four names.

## Decisions worth flagging

- **A clone of a service-agent box runs that agent.** The plan kept `agent: 'none'`, but then the per-box secrets had no create to ride into and nothing would ever carry them — the "second bot" was a directory. TUI boxes are unchanged.
- **The render is host-side, in `prepareClone`.** The plan preferred an in-box `run_once: { check }`; it cannot work — the only available check is "does SOUL.md exist" and the clone has just copied one, so it never fires.
- **`optional` was the vocabulary `required:` wanted.** Honoured on a create (a first bot has nobody to collide with), ignored on a clone (which *is* the collision). The refusal runs before the export, so nothing is left behind.
- **No `word: true` on `ReplaceRule`** — `regex: true` with `\bAda\b` already does it.
- The plan's "the hub loads no agent modules" fact was stale; corrected in the doc.

## Phase 4

A bot writes its own `identity` rule-set through a new `agentbox-identity` skill, nudged by a paragraph appended to the box facts *while* `agentbox.yaml` declares none. The yaml is the only state — the file is regenerated every boot, so the nudge disappears by itself and no marker can go stale. A supervisor task could not do this: it needs a judgement about which words are your name, on a turn.

## Verified live on docker, `ada` -> `bea`

- no `bea.env` -> refused, naming the path, with **no** dir/container/box left behind
- with it -> `bea` runs openclaw; own gateway token (`272e0674…` vs `402d973d…`); own 0600 `~/.openclaw/.env`; `SOUL.md`/`IDENTITY.md` say bea while **`Adafruit` survives**; `USER.md` regenerated; workspace at `<project>/.agentbox/bots/bea/workspace`; `config validate` clean and `/healthz` 200 on both; ada untouched
- ordinary create logs `carry: no …/ada.env — ada starts without it` and proceeds
- the nudge appears with the sentinel removed and is gone once restored, box facts intact

4,875 unit tests, typecheck and lint green; both new guards mutation-checked. Cloud `applyCarry` goes through the same call but no cloud clone has run — noted in the backlog.

---

> [!NOTE]
> **Medium Risk**
> Changes hub clone and box-create flows (agent selection, export, secrets refusal, yaml surgery); mistakes could break clones or leak shared channel tokens, though refusal-before-export and tests mitigate that.
> 
> **Overview**
> **`agentbox clone` now produces a real second bot** when the source runs a service agent (e.g. OpenClaw), not an agentless workspace copy. The clone still gets a fresh gateway identity and no copied credentials, but the create path runs the same agent and applies new **`clone`** spec behavior: drop regenerated files, render identity files, and require per-box secrets before export.
> 
> Agents declare **`clone.drop` / `clone.render` / `clone.perBoxCarry`** (replacing `workspaceArtifacts`). The hub **`prepareClone`** refuses early if the per-box env file is missing, exports with agent-scoped drops, then **`renderCloneIdentity`** rewrites `SOUL.md`/`IDENTITY.md` via the workspace `identity` rule-set and **`{{AGENTBOX_BOX_NAME}}`**, and strips those rules from the clone’s `agentbox.yaml` so a clone-of-a-clone can write its own rules. Default clone workspace for project boxes moves to **`<project>/.agentbox/bots/<name>/workspace`**.
> 
> OpenClaw ships the **`agentbox-identity`** skill (baked on all providers) plus a box-facts nudge until `agentbox.yaml` has identity rules. Create/clone carry paths merge **`withPerBoxCarry`** on docker and cloud. CLI/docs/changelog reflect the new contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96fbb6efc7beb4a6dadf9b7b17e370cedab6fc6c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

PR #376, opened 2026-09-08
feat(openclaw): open a bot's dashboard signed in, from the CLI, the hub and the tray

madarco:
Marco could not reach his Hetzner bot's Control UI: it asks for an auth token the gateway generated for **itself**, inside the box, and nothing handed it over. `agentbox openclaw url` printed the URL and the token on separate lines, leaving you to assemble `<url>/#token=…` by hand; the tray and the hub UI had no way to get it at all.

## What lands

- **`agentbox openclaw url`** prints a third line: `open: <url>/#token=…`. The bare URL stays on line one, because that is the documented `| head -1` surface and a fragment there would break appending a path.
- **`GET /api/v1/boxes/{id}/web`** → `{ url, signInUrl }`, so the hub UI and the tray get the same link. The token is never on the box list: reading it is an exec into the box, and it is a live credential.
- **The tray's Open Web** goes through that route for **every** box.
- **`agentbox url`** does too, instead of reading the box list's recorded URL for docker boxes.
- **A service agent can name a COMMAND for its own token** — openclaw now uses `openclaw dashboard --json --no-open` instead of us reading `~/.openclaw/openclaw.json`.

## Three things Marco pushed back on, each of which changed the design

1. **"Why not resolve at click time in general?"** The first cut carried a `serviceAgent` flag on the box list so clients knew when to ask — making a bot the special case. But `GET …/vnc` is *already* click-time for exactly this class of reason. The flag is gone and clients call `/web` unconditionally.
2. **"Why parse their config file?"** We were depending on openclaw's private layout; if they move that key we produce a link that silently opens to a login prompt. `dashboard --json` is the interface they ship. `urlFields[].command` runs it in the box, `fromUrlFragment` lifts the token out of the URL it prints, `fragmentKey` puts it into the fragment of the **host's** URL — only the fragment carries over, since the daemon's URL names its own loopback. `file` stays for a daemon with no such command.
3. **"Does the non-service path still read the cache?"** Partly, and worth stating: the hub payload is no longer the source, but for docker `resolveUrl` itself reads `box.portlessUrl` (the alias is stable, so that is correct). Verified by poisoning the record — the poisoned value came through. The fix is real for SSH-forward providers (hetzner/DO), a no-op for docker+Portless.

`--no-open` is load-bearing (the daemon would launch a browser *inside* the box) and `--yes` is deliberately absent (a read must never install or start the gateway). Both pinned by a test.

## Corrections made along the way

- I claimed a stale port was part of why it failed. **Wrong** — the old port still answered 200. The recorded port belongs to whatever forward existed when it was written and drifts across hub restarts (`61770 → 65217 → 63770 → 49870` while the list kept `61770`), but it is not dead. Wording corrected in code and docs.
- I claimed the config file was big. It is **1.3 KB**. The argument for the command is interface stability, not bytes.

## Verified live

Against the real Hetzner bot and a docker box: `agentbox openclaw url`, `agentbox url --print` (bare URL on stdout, sign-in link on stderr) and the hub route all return the same token, and the link answers 200. A plain docker box returns `{url, signInUrl: null}` — unchanged behaviour. 14 new unit tests (banner-before-JSON, one exec per source, command failure → falls back to the bare URL). Full suite, typecheck and lint green.

The tray half is on its own repo's `nightly` (`38bb357`, `c31e05c`), per its no-PR convention, and its `CLAUDE.md` records the new route.

---

> [!NOTE]
> **Medium Risk**
> Changes how web URLs and gateway tokens are resolved and exposed across CLI, hub API, and UI; token reads exec into running boxes but are on-demand, uncached, and best-effort with fallbacks to the bare URL.
> 
> **Overview**
> OpenClaw and other **service agents** can now open their Control UI **already signed in**, instead of printing a bare URL and a separate token you had to stitch into `#token=…` yourself.
> 
> **Shared resolution in `sandbox-core`:** `readServiceUrlFields`, `serviceSignInUrl`, and `serviceAgentForBox` centralize reading daemon tokens (via in-box **command** or JSON **file**) and building fragment URLs. OpenClaw switches to `openclaw dashboard --json --no-open` with `fromUrlFragment` / `fragmentKey` on `AgentServiceUrlField`, rather than reading `openclaw.json`.
> 
> **Click-time hub API:** New **`GET /api/v1/boxes/{id}/web`** returns `{ url, signInUrl }` (live `resolveUrl` + token exec, `no-store`). A matching **`/boxes/{id}/web`** page route **302**s to `signInUrl` so the hub **Open web** button works like VNC (sync navigation, no popup block). **`webOpenTarget`** gates the button on raw `state`, not stale list `webUrl`.
> 
> **CLI:** `agentbox url` uses the hub `/web` route for all providers (not docker’s cached `webUrl`), opens the sign-in link when present, and keeps `--print` pipeable. `agentbox openclaw url` / service launch adds an **`open:`** line after the bare URL.
> 
> **Clone safety:** Clone export uses `serviceAgentForBox` for agent scoping; **`agentCloneDropPaths()`** with no agent id now drops **nothing** (avoids stripping user `AGENTS.md` / `USER.md`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e3ae94cf869f77cc74a300f5bf2aa0573d735c4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #377, opened 2026-09-09
feat(prompts): ask create-time questions from the tray and the hub UI

madarco:
A box created from the tray or the hub web UI silently skipped every
host-boundary decision the CLI asks about. Two were live bugs:

- **`carry:`** — the gate lived in `apps/cli` and was called only by the three
  CLI create commands. The hub never parsed `carry:`; `POST /api/v1/boxes`
  accepted a pre-resolved `opts.carry` that only the CLI ever sent. A tray box
  got no carried files, no error, no log line, and `resyncCarryFiles` could not
  recover it (it returns early when `box.carry.entries` is empty).
- **`--model-auth`** — same shape. `GET /api/v1/agents` offers `openclaw`, so a
  tray-created OpenClaw box came up with no model provider. The queue plumbing
  for `borrowCredentials` already existed end to end, but `parseCreateBoxOpts`
  never parsed the field, so the key `hub-backend` reads was dead on the wire.

Both gates were TTY-only by construction: carry *threw* on a non-TTY,
model-auth silently resolved to `none`. Neither could reach a GUI.

## One mechanism, not two fixes

`PromptRequest` in `@agentbox/core` is deliberately two halves. The generic one
— `kind` / `title` / `body` / `choices` / `fallback` — renders and answers *any*
prompt, including a `topic` a client has never seen, so the hub can add a gate
without shipping new UI. `PromptDetail` is the opt-in half a client *may* draw
richly (`file-table`, `credential`); every variant carries a plain `summary`, so
an unrecognised one still renders. Ids are content-addressed over the question,
so an answer collected before `agentbox.yaml` changed no longer matches.

Both gates moved to `@agentbox/sandbox-core` and take a `PromptAsker` instead of
calling a prompt library. Three implementations drive them: clack in the CLI, a
**collecting** asker that *is* the hub's preflight, and a map-backed one
replaying the client's answers. The preflight cannot drift from what the create
asks, because it runs the same gates.

`POST /api/v1/projects/:id/create-preflight` returns the questions; answers ride
back as `opts.promptAnswers`. Gates run before the job is queued — the same
front-loading the CLI's `-i` path does — so **the queue worker is unchanged**.

CLI behaviour is preserved by its asker: `-y` still does not auto-approve carry,
and a non-TTY still refuses a `required` prompt with the gate's own hint.

## Verified against a live hub

- Preflight returns this repo's real 15-entry carry table with correct sizes and
  `—` for missing optional entries.
- Answering `approve` copied **13/15 entries** into a hub-created box at 0600
  owned by `vscode`; `skip-this-run` left them out. The answer changes behaviour.
- A create with **no answer**, and one with a **stale id**, are both refused.
- Tray models decode the live payload, and a simulated *future* detail variant
  falls back to its summary rather than failing the decode.

That testing caught a real bug: the collecting asker answers each prompt with
its fallback, and carry's fallback is `cancel` — so the carry gate aborted the
run and model-auth never appeared in the list. Fixed with an explicit dry-run
mode, plus a regression test.

## Note on scope

This branch also carries three **openclaw model-auth** commits it is built on
(`model-auth-gate.ts` exists only there), plus two unrelated fixes that were
already local. Happy to split if you'd rather review them separately.

The macOS tray half is in `madarco/agentbox-tray` on `nightly`.

## Known gap

The hub web modal is written, typechecks and lints, but could not be exercised:
every hub page 500s in the standalone build on a pre-existing untraced-`execa`
error that reproduces identically on the pre-branch staged bundle. Unrelated to
this change, but it means the web renderer is unverified.

---

> [!NOTE]
> **High Risk**
> Changes how host secrets and borrowed OAuth logins are approved and copied at create time across CLI, hub API, and UI; incorrect gate replay or validation could skip carry, mis-seed credentials, or refuse valid creates.
> 
> **Overview**
> Introduces a shared **create-time prompt** model (`PromptRequest` / `PromptAsker` in `@agentbox/core`) and moves **carry** and **model-auth** decisions into `@agentbox/sandbox-core`, so the CLI, hub backend, and web UI run the same gates instead of TTY-only CLI logic.
> 
> **Hub & API:** Adds `POST /projects/{id}/create-preflight` (collecting asker) and threads `opts.promptAnswers` plus parsed `borrowCredentials` through box create validation and `hub-backend` **before** the job is queued. The create-box modal runs preflight first, renders prompts via `PromptView`, and surfaces `unavailable` when the hub cannot read local project files or host logins.
> 
> **CLI:** Replaces local carry prompting with the shared gate + `clackAsker`; adds `--model-auth` on service agents and `resolveModelAuth` before box creation; propagates `borrowCredentials` through queued jobs. **OpenClaw** gains `modelAuth` registry settings (`openclaw.modelAuth`), an `openclaw-model-auth` ingest task, and credential fan-out for boxes that borrow Codex logins.
> 
> **Other:** `BoxRecord.borrowedCredentials` and clone/create plumbing; signed service URLs for VNC/web open paths; carry resolver import moved to sandbox-core; docs and OpenAPI updated for preflight and model auth.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c01c2d38a71b1dd6d61aa4ae9d59987bdc48b3a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
Superseded by #378, which is stacked on this branch and therefore contains all 13 of these commits plus the model-auth work. Closing so there is one review surface rather than two overlapping ones — the code here is not abandoned, it merges to nightly as part of #378.

PR #378, opened 2026-09-09
feat(model-auth): model-provider logins for every agent, not just openclaw

madarco:
Generalises model-auth from "openclaw borrows codex" to a source model any agent
can use, then closes the gaps that made it work only from the CLI.

**Supersedes #377** — this branch is stacked on `feat/create-prompts`, so merging
it brings both. #377 can be closed on merge.

## What changed

`AgentSyncSpec.modelAuth` becomes `{ sources, ingest?, promptOnCreate? }`:

- A **source** is `agent` (another agent's login FILE) or `env` (a provider API
  key in the host env). `modelAuthSourceId` keeps agent ids bare (`codex`) and
  prefixes env keys (`env:XAI_API_KEY`), so every existing `--model-auth codex`
  invocation and stored config value keeps working.
- An **ingest** is a `serviceTask` (openclaw's, in its own DAG) or a `command`
  the host runs at the launch seam. A TUI agent must use `command`: it has no
  supervisor unit, and ctl's wire cannot express a task without a service.
- **pi and opencode can now run on your Codex login.** Both already store a
  ChatGPT-subscription profile in the same five-field shape under the same OAuth
  client, so the import is a field mapping, not a plugin install.
- **Only openclaw asks** (`promptOnCreate`). Coding agents have their own
  sign-in; they keep the flag, the config key, the ingest and the fan-out, they
  just never prompt. The question is a one-pick list of the available logins plus
  "None", replacing a yes/no that hid *which* login it meant behind the word Yes.
- `--model-auth <source...>` reaches every declaring agent, and `enum-list`
  lets `<agent>.modelAuth` hold several.

## Two measurements that shaped the code

Taken before writing it, on real credential stores:

1. **`expires` must come from the access token's own `exp` claim.** Writing 0 or
   a past value reads as harmless — "it will just refresh" — but the consumer's
   refresh call REJECTS a codex-issued refresh token (`invalid_state`), verified
   back to back against a token that refreshed and rotated fine. The plan said to
   write it in the past; that would have shipped auth that never worked.
2. **A seeded box therefore cannot renew itself.** The credential fan-out
   re-pushing the host's login is the only renewal path, which is why every
   ingest gates on a hash of the SEED.

Claude's OAuth blob stays non-borrowable: a consumer's refresh rotates the token
and logs the host and every claude box out.

## Bugs found by the smoke test

- **The hub's docker queue worker never ran the ingest** — a tray- or web-created
  pi/opencode box got the login copied in and never imported it. Silently
  unauthenticated. `restoreAgentSessions` had the same hole. A source-level guard
  now scans for anything launching an agent session; it found a **fifth** seam on
  its first run (`dashboard.ts`, seven call sites).
- **The ingest marker ran once per HOST, not per box** — it lived in
  `~/.pi/agent/`, which is the SHARED `agentbox-pi-config` volume. Box #2 read
  box #1's hash, logged "already imported", and came up on the wrong credential
  while `pi auth check` still said ready. Moved to a per-box `/run/agentbox/`
  path. openclaw was never affected: its volume is per-box.
- **Every hub PAGE 500'd** (pre-existing, unrelated). `execa` in
  `serverExternalPackages` made turbopack emit an async external under a
  generated id naming no real package; API routes never load that chunk, which is
  why the hub looked half-alive.

## Verified

Every agent, every surface, with a real model turn as the pass condition — box
readiness proves nothing here.

| | openclaw | pi | opencode | claude | codex |
|---|---|---|---|---|---|
| preflight asks? | yes | no | no | no | no |
| CLI turn | `OCL_OK` | `PI_OK` | `OC_OK` | session live, no seed | session live, own login |
| Web | prompt renders | `WEB_PI_OK` via queue worker | — | — | — |
| Tray | decode verified vs live payload | — | — | — | — |

Tray also proven against a malformed `credential` detail: it used a plain `try`
over non-optional paths, so one bad detail killed the whole `PreflightResponse`
and created the box with no answers. It now degrades to its summary, and all
four prompts in the test payload still decode.

---

> [!NOTE]
> **High Risk**
> Changes how host model-provider credentials are chosen, copied, and ingested across CLI, queue worker, dashboard, and credential fan-out—mistakes surface as silent auth failures rather than obvious errors.
> 
> **Overview**
> Extends **model-auth** beyond OpenClaw: agents that declare `modelAuth` get `--model-auth`, a host-side **`resolveModelAuth` gate** before create, **`borrowCredentials`** on the box record, and **`runModelAuthIngest`** immediately before any agent session starts (foreground create, start/attach, session restore, queue worker, dashboard). **`agentbox credentials propagate`** also re-pushes borrowed logins and forces ingest for **every** agent on the box.
> 
> Stacks **create-time prompts** on the same shared gates: carry and model-auth logic live in **`@agentbox/sandbox-core`** with a **`PromptRequest` / `clackAsker`** CLI path; the hub adds **`POST …/create-preflight`**, **`runCreateGates`** + collecting/replay askers, **`opts.promptAnswers` / `carryYes`**, and the create-box modal **`PromptView`**. Clones inherit **`borrowCredentials`** and set **`carryYes: true`** so required carry does not block headless creates.
> 
> Smaller fixes: **service sign-in URLs** for VNC/in-box browser, **`enum-list`** validation for `--agent-setting`, **`cp`** copy helpers moved to sandbox-core, hub **`execa`** bundling (drops broken turbopack external), and docs for model-auth sources and API preflight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c632031e470e4def153d8c5f1600e2832529259. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

madarco:
bugbot run

PR #380, opened 2026-09-09
feat(hub): open the create pickers on what a project last used

madarco:
## What

Creating a box from the hub web UI or the tray always started on **docker + claude**, and the modal reset the provider to docker on every project switch. A project that lives on hetzner with codex had both re-picked every time.

Every create path now records the pair it used on the project's registry entry, the hub API surfaces it on `Project`, and both pickers open on it.

## How

- **Store** — `~/.agentbox/projects/<hash>/meta.json` gains `lastProvider` / `lastAgent` / `lastUsedAt`, written by a new `recordProjectLastUsed()` in `@agentbox/config`. In the registry rather than derived from live boxes, because the memory has to outlive destroying the last box — which is exactly when the next one gets created. **Not config**: `loadEffectiveConfig` never sees it, so CLI behaviour is unchanged and `box.provider` stays the explicit, user-authored answer there.
- **Update-only** — it refuses to create a registration, so a control box's per-job clone (or any unregistered path a create resolves) can never mint a project card.
- **Write seams** — the hub's `create()` past its gates (web, tray, CLI via the hub, `--via-hub`), `agentbox create`, and `agentbox <agent>` above all five of its branches. Recorded at enqueue: it is a UI default, and the provider just picked is the right pre-selection even if that build fails. An agentless create records only the provider.
- **Read** — `Project.lastProvider/lastAgent/lastUsedAt`. A project registered before this feature falls back to its newest box; the hosted/control-box path derives the same from its newest registration. `Project.provider` is untouched (`computeNeedsSetup` reads it).
- **Clamping** — new pure `apps/hub/lib/boxes/project-defaults.ts`. A remembered provider can be unbaked, owned by a control box, or gone; a remembered agent can be an uninstalled plugin. Both fall back to docker / the first agent.
- **Bug fixed on the way**: `touchProjectMeta` rebuilt `meta.json` from scratch, so the hub's per-poll `registerProject` would have erased the new fields seconds after they were written. It now merges, and takes the file lock it shares with the new writer.
- **Tray** (madarco/agentbox-tray@3428680) — same defaults; a remembered agent sitting in the overflow is promoted into the segment row so it is selectable at all.

## Verified

- `pnpm build`, `typecheck`, `lint`, `test` green; new tests in `packages/config/test/meta-json.test.ts` and `apps/hub/test/project-defaults.test.ts`.
- Live: a real daytona+codex create through `POST /api/v1/boxes` recorded the pair, it survived a minute of hub polling (the clobber fix) **and** survived destroying the box; the modal then opened on Daytona/Codex for that project, Hetzner/Codex for another, Docker/Claude for one with no memory; an explicit pick survived 6s of SSE re-renders.

https://claude.ai/code/session_01WQzzMHt7KZy4Bmv8gF1EoS

---

> [!NOTE]
> **High Risk**
> Broad changes to create routing, credential borrowing, bot identity backup/restore, and hub project metadata—any mistake can leak secrets, duplicate live bot identities, or break creates across providers.
> 
> **Overview**
> **Create pickers** now remember what each project last used: `recordProjectLastUsed()` writes `lastProvider` / `lastAgent` / `lastUsedAt` into the project registry from `agentbox create` and `agentbox <agent>` (after refusals, before routing), and the hub exposes those fields with **`project-defaults`** clamping so stale or unavailable providers/agents fall back to docker/claude instead of failing creates. **`touchProjectMeta`** is fixed to merge rather than wipe that metadata on hub polling.
> 
> The same release adds a **service-agent CLI** (`surface: service`, e.g. OpenClaw): create-or-resume, poll ctl until `ready_when` passes, signed **url** output, and subcommands for status/logs/restart/stop—separate from the TUI attach path. **Bot backup/restore/clone** land via `download --backup`, `--restore`, hub **`agentbox clone`**, plus an **`agentbox-identity`** skill for clone-time name substitution rules.
> 
> **Model-auth seeding** (`--model-auth`, `borrowCredentials`) runs a host gate before create and **`runModelAuthIngest`** on docker session start/resume and queue workers so borrowed logins refresh. **`--persistent`** / **`box.persistent`** propagate through create, hub jobs, and clone with provider caps refused up front. Smaller fixes include agent **run-env** on VPS providers, excluding host **`.agentbox/`** from workspace seed, and opening approval **links** on the answering machine.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 331190a7158fe6e5319ef384d9690c9e4c9345dd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
CI was red on `@agentbox/ctl`'s `gh pr comment consumes every value-taking flag gh declares` — unrelated to this feature, but a real bug it caught: gh 2.99 added `--attach <file>` to `pr comment`, and the shim's value-flag map didn't have it, so the file path was read as the PR ref (no branch injection → the host resolves the comment against its own checkout, i.e. a comment on the wrong PR). Fixed in 036ad0ca.

madarco:
Addressed the `/code-review medium` findings in 625e743b:

- **Modal re-seeded under the user (medium)** — the seeding effects keyed on values recomputed from `projects`, which the store replaces on every SSE tick, so a create recorded elsewhere for the same project flipped both pickers mid-form. Seeding is now latched on the selected project id; the two pin flags are gone (the latch subsumes them). Verified live: with the modal open and untouched, an out-of-band record of a different pair left the form on `daytona/codex` through 9s of polling, and a fresh load then opened on the new pair.
- **`agentbox create` recorded above the refusal gates (low)** — moved below them, matching the invariant `create-action.ts` already states.
- **`kind` fallback emitted `lastProvider: 'cloud'` (low)** — now says nothing instead; the agent goes through the one queue wire spelling (`claude-code` → `claude`) with anything else passing through, so a service/plugin agent isn't dropped. Test updated (it had locked in the `'cloud'` behaviour).
- **15s stale-lock window on a hot-path file (low)** — `withFileLock` takes optional stale/acquire windows; `meta.json` uses short ones, since `registerProject` runs for every box root on every dashboard poll.

PR #383, opened 2026-09-10
feat(carry): approve a project's file list once, not on every create

madarco:
## What

Starting three boxes for one project meant approving the identical carry table three times — in the terminal, in the hub web modal, and in the tray card. Nothing remembered the answer anywhere.

The host now records the approved file list per project. While that list is unchanged, every surface creates boxes without asking.

## How

`tools:` in `agentbox.yaml` only **requests** host access; the host approves once and the grant lands in a per-project `tools.yaml`. `carry:` has the identical trust split — a committed block requesting host files — but no grant store, so it asked forever. It now gets the same treatment: `~/.agentbox/projects/<hash>/carry.yaml`, holding the approved list (readable, not an opaque digest), never the repo.

**Keyed on the list, not on contents.** The existing prompt id could not be reused: `buildCarryPrompt` includes each row's `bytes` — right as the anti-replay key for a single create, wrong for a standing approval, since editing an approved `secrets.env` would re-ask forever. `carryGrantId` hashes the same rows minus `bytes`, keeping `src`/`dest`/`kind`/`mode`/`user`/`flags`/`warn`: shape and safety, not content. A file that becomes a folder, changes mode, or starts resolving through a symlink out of `$HOME` still re-asks.

**One seam.** The check lives in the shared gate that the CLI, the hub preflight and the create replay all run, so all three surfaces inherit it from one edit — and because the preflight runs it too, a granted project shows the web and tray clients no question at all. **No API, schema or client change in either app.**

- The gate takes the grant as an argument rather than reading it: `packages/sandbox-core/test/` has no HOME isolation, so a gate touching `~/.agentbox` would write to the real home. Writes happen only at create intent (the CLI wrapper; `hub-backend.create`, never the preflight).
- `--carry-yes` / `opts.carryYes` approve the run but never leave a standing grant — a scripted bypass is not a human reading the table. (That is also what `clone`/`restore` pass.)
- `--carry ask` re-opens a granted list. It had to stop being commander's **default** value for that to mean anything — otherwise every run looked like an explicit `--carry ask` and the grant would never apply. That is the golden CLI-surface fixture diff.
- `apps/hub` gets the temp-HOME test isolation `config` and `relay` already have: these tests read host state through `loadEffectiveConfig`, so a developer's own `~/.agentbox` could change what they assert.

## Verified

`pnpm build`, `typecheck`, `lint`, `test` green; new tests in `packages/sandbox-core`, `packages/config` and `apps/hub`.

Live, against the built code and a running hub:

| | asked? |
| --- | --- |
| first create, no grant | yes |
| same list again | **no** — `carry: approved earlier for this project (1 file, list unchanged)` |
| carried file's **contents + size** changed | **no** (the point of the feature) |
| `dest` changed in `agentbox.yaml` | yes |
| `--carry ask` | yes |

Through the hub: `POST /projects/{id}/create-preflight` returned `prompts: []` for a granted list, `['carry']` after the list changed, and a real docker create with an approved answer wrote the grant and landed the file at `~/.moved.env` inside the box.

https://claude.ai/code/session_01WQzzMHt7KZy4Bmv8gF1EoS

---

> [!NOTE]
> **High Risk**
> Changes span credential seeding (`borrowCredentials`, carry grants), bot state backup/restore/clone, hub create/clone APIs, and multi-path create flows (CLI, queue, service agents)—any bug can leak host files or overwrite bot identity.
> 
> **Overview**
> This PR is a large product slice around **service bots (OpenClaw)** and **host trust**, not only carry.
> 
> **Carry:** The host can **remember an approved `carry:` file list per project** (grant keyed on list shape, not file bytes). Repeat creates skip the prompt when the list is unchanged; `--carry ask` forces re-approval; `--carry-yes` stays a one-shot bypass. The shared carry gate is wired through CLI, hub preflight, and create replay; `carry-prompt.ts` is removed in favor of the shared path.
> 
> **Service agents:** **`surface: 'service'` agents** get a separate Commander tree (`service-factory` / `service-action`): create-or-resume, poll supervisor readiness (not “running” before probes pass), print signed-in URLs, plus `status` / `logs` / `restart` / `stop` / `url` (and optional `attach` for a declared REPL). Registry rows drive registration; TUI agents keep `attachWrapped`.
> 
> **Bot backup / restore / clone:** **`agentbox download --backup`** and **`--restore`** (workspace on `create`, full state on `agentbox <agent> --restore`), **`agentbox clone`** via hub `POST …/clone`, plus **`agentbox-identity`** skill and runtime staging for clone identity rules in `agentbox.yaml`.
> 
> **Model auth & persistence:** **`--model-auth`** and **`borrowCredentials`** on create, with **`runModelAuthIngest`** before docker/queued sessions start; **`--persistent` / `--no-persistent`** with provider refusal where the platform caps sessions. Create paths record **last-used provider/agent** for hub/tray pickers.
> 
> **Hub / URLs:** Live resolution for box web URLs (`GET …/boxes/{id}/web`, `agentbox url` via hub), approval answers that **open links on the host**, and related docs/CHANGELOG entries. **Fixes:** agent **run-env** reaches VPS units; workspace seed **excludes `.agentbox/`**; enum-list agent settings parsing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d286f3a0ab392cce0848b06927f633e664d35607. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Addressed all seven `/review medium` findings in 526c389c. Five were ways a standing grant could approve a copy nobody agreed to:

1. **Digest holes (the serious one).** `exclude:` / `replaceEnvs` / `replace` / `rules` were not hashed — and `bytes` was the only field that would have moved. Editing a committed `agentbox.yaml` to drop `exclude: [gh-token]` from `~/.config` kept the same grant id and silently widened the copy to include the token. All four are hashed now (`rules` expands into `replace` at resolve time, so two keys cover the pair). Verified live: dropping an exclude rotates the id.
2. **Unvalidated `--carry <mode>`.** With the commander default gone, `--carry skpi` was indistinguishable from "no flag" and fell into the granted-silent path. Parsed once and refused, as `fork` already does.
3. **No revoke.** `--carry ask` re-opened the question but answering Skip left the grant in place. The gate now reports whether a human was actually shown the table; both callers withdraw the grant on a decline, while a per-run `--carry skip` (never seen by a human) leaves it alone. Verified live both ways.
4. **`AGENTBOX_CARRY=skip` beat `--carry ask`** (`flag ?? env` fell through), inverting CLI > env. Now a pure `resolveCarryFlags`, unit-tested.
5. **Hub had no `carryAsk`**, so from the web UI and the tray a standing approval was permanently invisible and unrevokable. Accepted on both the preflight and the create now.

And the two doc corrections: the page promised a non-TTY create fails loud for a non-empty `carry:` block (a grant makes that false), and never said a grant spans providers — a list approved for a local docker box is carried to a cloud box too. Both stated explicitly.

Finding 7 (provider scope) was a deliberate choice, but it was undocumented — that is what the second doc change fixes.

madarco:
Landed on `nightly` as 526c389c (fast-forward, not squashed to main) — closing per the nightly integration flow.

PR #386, opened 2026-09-11
feat(sandbox-providers): add CreateOS sandbox provider

ashwaq06:
## What this changes

Adds `createos` as a sandbox provider, alongside the existing docker, daytona, hetzner, vercel, e2b, digitalocean and remote-docker backends. CreateOS runs Firecracker microVMs that pause to zero and resume by id, so a box keeps its workspace while idle without being billed for it.

The provider composes the shared cloud scaffold and overrides `buildAttach` only, because CreateOS has no AgentBox-managed SSH transport and attaches over a managed PTY instead. It needs no bake: valid credentials make it usable, and each create installs the runtime into a fresh rootfs.

Registered through the `PROVIDERS` table, so the CLI, hub and settings UI pick it up from one row.

## How it was tested

Rebased onto current main first, then on a clean checkout:

```
pnpm install --frozen-lockfile
pnpm lint        26/26 tasks
pnpm build       27/27 tasks
pnpm typecheck   49/49 tasks
pnpm test        48/48 tasks
```

Note that `pnpm test` needs `pnpm build` to have run first: `apps/cli/test/cloud-e2e.test.ts` resolves `../dist/index.js` while the describe body is evaluated, so on a tree that has never been built it fails to collect before `skipIf` can skip it. That happens on main too, without this branch.

`packages/sandbox-createos` also ships unit tests for size parsing, sandbox name capping, the build-attach argv and the review findings. The live test is opt-in and needs a CreateOS API key.

## Demo

`agentbox createos opencode` opening OpenCode inside a CreateOS sandbox: the provider creates the box, installs the runtime into a fresh rootfs, and seeds the workspace from a shallow clone.

https://github.com/user-attachments/assets/c31ba3be-98a8-4ddb-aec8-8de96789b5ef

The provider registered and selectable, the descriptor the table serves, and the checks above running green.

agentbox-createos

## Checklist

- [x] `pnpm lint && pnpm build && pnpm typecheck && pnpm test` passes
- [x] Public docs updated if a command, flag, config key, or default changed (`apps/web/content/docs/`)
- [x] I have read and agree to the [Contributor License Agreement](https://github.com/madarco/agentbox/blob/main/.github/CLA.md)


ashwaq06:
I have read the CLA Document and I hereby sign the CLA

madarco:
thanks @ashwaq06, great work! However the right approach is to create a plugin: https://github.com/madarco/agentbox/blob/main/CONTRIBUTING.md

https://agent-box.sh/docs/build-a-provider

BhautikChudasama:
hi @madarco can we use plugin to give similar experience like e2b / hetzner?

madarco:
> hi @madarco can we use plugin to give similar experience like e2b / hetzner?

yes is exactly the same. all providers will move to plugins eventually

BhautikChudasama:
thank you @madarco :)
lets close this PR @ashwaq06

PR #16, opened 2026-05-28
feat(vercel): add Vercel Sandbox provider

madarco:
## Summary
- Adds `@agentbox/sandbox-vercel` (`--provider vercel`), a fourth backend on top of `@vercel/sandbox` (Firecracker microVMs + snapshots), composed via the shared `CloudBackend` scaffolding like Daytona/Hetzner.
- Base env is baked once via `agentbox prepare --provider vercel` (Vercel can't build from a Dockerfile): boot `node24` → run `provision.sh` (AL2023: dnf deps, `vscode` user, agentbox-ctl + shims, Claude native installer, codex/opencode) → `sandbox.snapshot()`.
- Persistent sandboxes give pause/resume for free (`stop` auto-snapshots, `get({resume:true})` resumes); `destroy` purges the box's snapshot. Public HTTPS preview URLs via `sandbox.domain(port)`.
- No in-box `dockerd` (Vercel blocks nested containers) and no SSH — interactive attach is a custom `attach-helper.js` tmux bridge (`send-keys`/`capture-pane`) over the SDK.
- Checkpoints store the Vercel snapshot **id** in the cloud-checkpoint manifest (Vercel snapshots are id-addressed).
- OIDC auth decodes the JWT's `owner_id`/`project_id` and passes explicit creds — the SDK's env-OIDC path needs Vercel-CLI linkage a box doesn't have (found via live testing; also detects expired tokens with a clear message).

## Wiring
- Provider registry + argv-prefix sugar + `agentbox vercel` CLI command.
- Config: `ProviderKind` + `box.defaultCheckpointVercel` (+ JSON schema + checkpoint key resolution).
- Relay `resolveCloudBackend` learns `vercel`.
- `launchDockerd` opt-out added to the cloud scaffold.
- apps/cli bundles the SDK as `external` (undici uses dynamic require, which breaks the ESM bundle otherwise).

## Test plan
- [x] `pnpm build && pnpm lint && pnpm typecheck && pnpm test` all green (new unit tests: env-loader, credentials incl. OIDC decode/expiry, prepared-state, backend with mocked SDK, build-attach).
- [x] CLI smoke: `vercel --help`, `vercel login --status`, `create --provider vercel` routes correctly and surfaces the prepare gate; auth path reaches the Vercel API (verified live — blocked only by an expired OIDC token).
- [ ] Full live e2e (prepare → create → attach → checkpoint) pending a non-expired Vercel credential. Live-verify checklist in `docs/vercel-backlog.md`.

---

[...]

PR #389, opened 2026-09-17
feat(hub): narrow a workspace timeline to one manager session

madarco:
A workspace has many manager sessions, and the Timeline showed every one of them interleaved on a single branch graph. There was no way to ask "what did *this* session do".

## The API

`?managerId=` on `GET /workspaces/{id}/timeline` keeps only that session's rows, narrowing `items`, `live` and `summary` alike.

A row is the session's when it **stamped** it, **or** when it sits on a box (or create job) it **owns**. The second half is load-bearing: `managerId` is only written when `managerIdForTarget` could resolve one as the row was recorded, so most `box.*`, `git.push` and `pr.*` rows carry none. The box *lane* is checked alongside `boxId`, because a push or PR row reaches a box's lane through the branch carrier without ever naming the box.

An id no manager in the workspace has reads as an empty timeline, not a `400` — an unknown session is one that did nothing here.

### Lanes are deliberately untouched

The filter runs **after** `assignLanes` and **before** paging, so lane ids and forks are the ones the whole log gives. A kept row can therefore fork from or merge into a lane with no rows left on the page — which clients already handle, because paging could always strip a lane's own rows (`TimelineGraph.extend` draws exactly that case today). Re-deriving lanes on the narrowed set would have been the bug.

`buildTimelineSummary` was widened past `TimelineEvent` so the summary counts the rows actually on screen; aggregation drops nothing it counts.

## Verification

- Hub tests **358/358** (`timeline.test.ts` 27 → 31), typecheck 47/47, eslint clean.
- Live hub against a real workspace (81 rows, 2 sessions): **81 → 49 and 20**, each read showing only its own manager; unknown id → `{items: 0, live: 0}`.
- **Lane stability proved on real data:** all 49 kept rows have lanes byte-identical to the unfiltered read.

Two caveats worth carrying into review:

- The box-attribution clause is covered by unit test but **not** live here — every manager record on this machine has `boxIds: []` (those boxes were destroyed, and `reconcileManagers` drops dead boxes by design), so the live runs exercised stamped-row attribution only.
- `timeline.test.ts:348` is **flaky before this change**: it orders two `manager.note` rows sharing a timestamp, and `newestFirst` tie-breaks on a random id suffix. Not touched here; worth a stable tie-break separately.

## Companion change

The macOS tray gains a "This session" toggle in `WorkspacePaneHeader` that drives this parameter and revives `managerTaskFilter` (pinned to `false` since its checkbox was dropped), so one control narrows both the task list and the Timeline. That lives in `agentbox-tray` and is committed on its `nightly` — not pushed yet.

Docs updated in the same change: `api.mdx`, the OpenAPI query parameter, and the workspaces backlog.

https://claude.ai/code/session_013ZMo7ttmSSrYadHF9VrK4m

---

> [!NOTE]
> **Medium Risk**
> Timeline read path gains non-trivial attribution logic (box/job scope, reconciliation vs log); wrong rules would hide or leak another session's rows, but behavior is heavily unit-tested.
> 
> **Overview**
> Adds **`?managerId=`** on `GET /workspaces/{id}/timeline` so clients can show one manager session instead of the full interleaved graph. The filter applies to **`items`**, **`live`**, and **`summary`**; an unknown id returns an empty timeline (not `400`).
> 
> Rows are kept when the session **stamped** them or they belong to a **box/create job** it owns (`managerScope` / `rowBelongsToManager`), including unstamped `git.push`/`pr.*` rows via box id, box lane, or job key. Box ownership is inferred from `box.created`/`box.ready`/`box.failed` plus job↔box joins (not from incidental stamped rows like `task.assigned` onto someone else's box). Filtering runs **after** `assignLanes` and **before** paging so lane ids stay stable on the full log.
> 
> `buildTimelineSummary` now counts the narrowed row set. OpenAPI, `api.mdx`, and backlog item #16 document the parameter; hub tests cover multi-session, live rows, lanes, destroyed boxes, and job attribution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05e3e41c45207f47941cbe8289d26254319573c5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
bugbot run

PR #390, opened 2026-09-17
feat(agent): survive a hub restart, keep stdout clean, wait on many boxes

madarco:
Three fixes to the `agentbox agent` orchestration surface, all from driving a real fleet while the relay was restarted underneath it.

## 1. A hub blip no longer ends a wait

`agent wait-for` polled with no retry: anything that wasn't `not_found` was rethrown into `handleLifecycleError`, which printed a bare `fetch failed` and exited 1 — the same code as a legitimate timeout. A `hub restart` therefore ended a 50-minute wait after one failed poll, and the caller couldn't tell "the relay went away" from "the agent never got there".

The wait loop now lives in `apps/cli/src/lib/wait/agent-wait.ts` with every dependency injected:

- transient failures back off (250ms → 5s, jittered) and retry for the whole wall clock;
- after 3 consecutive failures the box's hub is **re-resolved**, which restarts a local hub that actually died;
- a hub that can't be brought up is *not* fatal — it's the case the caller asked us to sit through;
- `unauthorized` is fatal (exit 3); `not_found` is fatal immediately for a box that never answered, and after a short grace window (5 reads / 15s) for one that had — a hub's registry briefly 404s after a restart, but a box destroyed mid-wait must still be reported as gone;
- one stderr notice when a hub goes away, one when it comes back. Never per tick.

New exit code **7 = the hub never answered**, distinct from 1 (timed out while the hub was answering), 2 (unknown state/box) and 3 (credential rejected). `--json` failure envelopes carry `"unreachable": true` so a JSON caller needn't read `$?`.

`getAgentState` also takes a per-request `timeoutMs` (10s from the wait loop): without one a half-open socket blocked a tick forever, so the loop never re-checked its own deadline.

## 2. stdout carries only the value

`@clack/prompts` writes `log.*`, spinner frames and the raw cursor escapes from `@clack/core`'s `block()` to **stdout**, with no way to redirect them in 0.9.1. `agent state` reaches a spinner through `resolveBoxPromptSource` → `resolveHubApiTarget` → `autostartLocalHub()` — which fires exactly when the hub was just restarted. So `$(agentbox agent state X | tail -1)` returned `\x1b[?25h` followed by `working`, and every string comparison against it failed.

`withCleanStdout` (`apps/cli/src/lib/machine-output.ts`) diverts stdout to stderr for the duration of an action and hands it an `emit` bound to the real stdout. Applied to all five `agent` subcommands. A human loses nothing — stderr is still the terminal.

Also: emit-then-`process.exitCode` instead of `process.exit()` after a write, which was truncating `--json` payloads under a pipe.

## 3. Multi-box `wait-for`, and batch `destroy`

- `agentbox agent wait-for input-needed --box a --box b --box c` — repeatable, first match wins, prints `<box>\t<state>` (JSON: `{box,state,agent}`). A single-box wait is byte-for-byte unchanged. One box's hub being down doesn't end the wait for the others.
- `agentbox destroy --box a --box b` — every ref resolves before anything is destroyed (a typo fails the batch instead of destroying its neighbours), one confirmation for the set, sequential teardown, and the exit code reports the worst outcome.

The one-shot readers (`state`, `get-plan-question`, `approve`'s race guard) also retry a transient failure 3× before reporting — deliberately not pushed into `agentClaudeFrom`, which is `approvals --wait`'s inner poll.

## Verified live

Against a real docker box, killing the hub listener mid-wait:

- `agent state <box> | cat -v` prints `working` with no escapes, including on the autostart path (hub freshly stopped).
- Wait survived a hub kill: `note: hub unreachable (fetch failed) — retrying` → autostart → `note: hub reachable again — still waiting` → correct exit 1 for the genuine timeout that followed.
- With :8787 squatted so the hub could not be reclaimed: `{"matched":false,"elapsedMs":25040,"unreachable":true,"error":"fetch failed"}`, **rc=7**.
- Multi-box: `waitsmoke<TAB>prompt`, and the JSON envelope; single-box output unchanged.
- `destroy --box waitsmoke --box twin -y` → two `destroyed …` lines, `destroyed 2/2` on stderr, rc=0.

Plus 15 new unit tests for the wait engine (fake clock), 3 for the stdout boundary, 5 for the batch exit code; full `pnpm test` / `typecheck` / `lint` / `check:plugin-skill` green.

## Docs

`background-and-parallel.mdx` (multi-box example + exit-code table), `cli.mdx` (both `--box` flags, exit codes), and the `agentbox-info` host skill — whose written contract line said "exit 0 on match, exit 1 on timeout" — rewritten with the real table, the multi-box waiter, and a fan-out recipe that reads the captured value directly.

## Follow-ups (not in this PR)

- `queue wait-for` has the same "any error → exit 1" weakness.
- `wait-for` still polls; `/api/v1/boxes/:id/stream` carries `box-status` and could drive it (the retry logic would still be needed around it).
- Unrelated, found while testing: `agentbox create -y --carry-yes` (and `AGENTBOX_CARRY_YES=1` / `--carry skip`) fails non-interactively with "Waiting on an answer: Copy these files into the box?" — reproduced on the released 0.32.0 CLI, so pre-existing.

---

> [!NOTE]
> **Medium Risk**
> Changes exit codes and stdout contracts for `agent` scripting paths and batch destroy; behavior is intentional but scripts that assumed exit 1 for all failures or dirty captured stdout may need updates.
> 
> **Overview**
> Hardens **agent orchestration** for scripting and multi-box fan-outs: waits survive hub restarts, stdout is capture-safe, and several boxes can be waited on or destroyed in one command.
> 
> **`agent wait-for`** moves polling into `runAgentWait` with backoff, hub re-resolve after transient failures, per-read `timeoutMs` on `getAgentState`, and a new exit **7** when the hub never answers (vs **1** timeout while the hub was up). Repeatable **`--box`** races multiple boxes; multi-box output is `box<TAB>state` (JSON wraps `box`, `state`, `agent`). One-shot **`state`** / **`get-plan-question`** / approve race-guard reads retry briefly on blips.
> 
> **`withCleanStdout`** wraps all `agent` subcommands so only the data line hits stdout; spinners and logs go to stderr. JSON paths use **`process.exitCode`** instead of **`process.exit`** after writes so pipes don’t truncate payloads.
> 
> **`destroy`** gains repeatable **`--box`**: resolve all refs before tearing anything down, one confirmation, sequential teardown, batch exit via `decideDestroyBatch` (worst outcome wins).
> 
> Shared **`collect`** helper for repeatable `--box`; docs and the **agentbox-info** skill document exit codes, multi-box recipes, and batch destroy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4138a86efb49f469d6f10e93d1e519926e797acb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
Fixed all three findings in 7adf6715 — they were real:

1. **Wait exits before retrying hub** — confirmed. The resolve loop treated an unreachable hub as terminal, so a wait started a second after `hub restart` got an instant exit 7 (contradicting what 7 documents), dropped the other `--box` targets, and emitted no `--json` envelope. Losing the hub *mid*-wait was handled; losing it one second earlier was not. Targets now enter the loop without a source and `onStale` re-resolves. Only `unauthorized` stays fatal.
2. **Destroy overwrites hub exit codes** — confirmed, and it was a regression for the single-box case too. Precedence is now a pure `resolveDestroyExit(outcomes, hubCode)`.
3. **Batch destroy reaps before resolving** — confirmed. `findOrphanContainer` is read-only now; removal happens with the rest of the teardown, after the confirmation, and orphans are listed in it.

6 new unit tests; verified live that an orphan survives a bad ref in the same batch and is still reaped on its own.

bugbot run

PR #392, opened 2026-09-19
fix(manager): a control box stops lying about the managers on your machine

madarco:
Auditing the manager path end to end against a real control box (the deployed Hetzner hub) for the tray work turned up five ways a hub that only HOLDS a manager's record answered for it wrongly. Each one is something a GUI renders as a fact, so each was fixed rather than filed.

- **`GET /workspaces` contradicted `GET /managers`.** `managerCounts` derived `managers.running` without looking at the heartbeat, so a control box reported a manager running for up to 30 minutes after the manager list correctly said `stopped` — two routes the tray reads in the same tick, disagreeing. It also made `removeWorkspace`'s guard refuse `workspace rm` for just as long. `managerStatus` now prefers a fresh report over the last-seen window, which fixes every caller at once.
- **A heartbeat never advanced `lastSeenAt`**, freezing the fallback clock at detect time: a manager alive for hours read `stopped` the moment one report went missing, and `sortViews` put the longest-running one last. A running report now stamps it; a stopped one deliberately does not.
- **A hub whose store is remote repeated that store's heartbeat-derived status for managers running on its OWN machine**, where the tmux server is readable. That sent `manager attach` to `resume`, and `resume` back to "already running; attach to it instead" — a closed loop. Own-machine rows are re-probed locally, keeping what only the holding hub knows (workspace name, task counts).
- **A `wrong_host` about a WORKSPACE named `Object.keys(hosts)[0]`**, so on a workspace mapped from two PCs the other one saw a host that was not its own and skipped the retry against its own hub. Those refusals now carry `details.hosts` as well, and the CLI (and the tray) check membership.
- **`POST /managers/{id}/notes` dropped `X-AgentBox-Session-Turn`** — the only turn reader when the transcript is on another machine. It is honoured when the caller is the manager the note names, and dropped otherwise.

Backlog item 22 records what is deliberately left: the heartbeat poster lives in the PC's hub daemon and nothing on a manager read path starts it once a control box is configured, so an `external` manager's liveness is still a window rather than a probe. Making it truthful is a decision about what a PC still runs once it has a control box, not a bug fix.

Also here: the `Error` schema in `openapi.ts` now documents `details` and the full code enum, `api.mdx` documents the two 409 codes and the retry rule, and three stale openapi descriptions (`hub`→`tmux`, the folder-based workspace prose, a `required` list the parser does not enforce) are corrected.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm lint` green. Six new unit tests, each mutation-checked (reverting the fix fails the test): the two routes agreeing after a stopped heartbeat, `lastSeenAt` advancing on a running one and NOT on a stopped one, the own-machine re-probe (and that a manager on another machine is left exactly as the holding hub rendered it), the `hosts` list, and the note's turn.

Live against the deployed control box at `116.203.220.73.sslip.io`, with the macOS tray pointed at it: the workspace, its timeline and its manager render; `GET …/managers/sessions` 409s there with `details.host` naming this Mac and the client's retry against its own hub answers 200; a manager registered as running on `tower` renders with every local affordance disabled.

The matching tray changes are `madarco/agentbox-tray@e147cad` (on its `nightly`).

---

> [!NOTE]
> **Medium Risk**
> Changes manager status derivation, heartbeat side effects, and host-routing retries across CLI and hub—behavioral fixes with broad API/CLI impact but covered by new tests.
> 
> **Overview**
> Fixes several ways a control box (or any hub that only **holds** records) misreported manager liveness and blocked clients from retrying on the right machine.
> 
> **Liveness and consistency:** `@agentbox/relay` now owns heartbeat staleness (`freshHeartbeat`, shared intervals), and `managerStatus` prefers a fresh heartbeat over the `lastSeenAt` window—so `GET /workspaces` manager counts align with `GET /managers`. Running heartbeats advance `lastSeenAt`; stopped ones do not.
> 
> **Local re-probe:** When the record store is remote, managers on **this** machine are re-probed via tmux/daemon instead of repeating the holding hub’s heartbeat-derived status (fixes attach/resume deadlocks).
> 
> **Multi-machine workspaces:** `wrong_host` refusals for workspace ops (start, sessions) can include `details.hosts`; the CLI’s `retryOnLocalHub` retries when this hostname is in that list, not only when it matches lone `details.host`.
> 
> **Notes:** `POST /managers/{id}/notes` passes timeline session meta so a manager’s own `agentbox manager note` can stamp the correct turn from `X-AgentBox-Session`.
> 
> Docs/OpenAPI describe error codes, `details.host`/`hosts`, and retry rules; tests cover the above behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60e832ab98a18861bd43dc0d8d366ab212318f6c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

madarco:
Integrated into `nightly` by fast-forward (`9bc118fe..1b82dd73`), which is why GitHub marked this merged on its own.

Landed on green CI **without a Bugbot review**: Bugbot answered `usage limit reached` on both the automatic run at open and an explicit `bugbot run`. Worth re-reviewing once the Cursor limit resets.

PR #393, opened 2026-09-20
feat(manager): run manager sessions on an AgentBox pty host, not tmux

madarco:
A hub-run manager lived in the user's **own** tmux server, so `manager.ts` could
only set session-scoped options — and every option that would fix the ergonomics
(`extended-keys`, `set-clipboard`/OSC52, `allow-passthrough`, the wheel rebind)
is a **server** option that would leak into every unrelated tmux session. That
is why the modified Enter never reached the agent (the tray rewrote Ctrl+Enter
to Ctrl+J to compensate), why the wheel dropped into copy mode instead of the
terminal's own scrollback, and why selection and clipboard went through tmux.

A manager now runs on an AgentBox **pty host**: one detached process per
session, holding the agent's terminal and handing raw bytes to whatever
attaches. Nothing sits between the agent and the terminal you look at.

**Measured live** against a real Claude Code manager: `ESC[13;2u` (Shift+Enter)
reached the agent unchanged and inserted a newline — two lines in the prompt,
nothing submitted, nothing rewriting it anywhere.

## What it is

- **`pty-host`** (`apps/cli/dist/pty-host.js`, a third tsup entry) owns node-pty
  and serves `~/.agentbox/pty/<id>.sock`. It outlives the hub — which restarts on
  every update — and the hub never requires node-pty: it spawns the entry and
  talks `node:net`. The spawn spec rides **fd 3**, not argv, so the session token
  and the launch script stay out of `ps` (the tmux carrier leaked the script).
- **Length-prefixed binary framing** (`packages/core/src/pty-protocol.ts`): pty
  output is arbitrary bytes and byte-exactness is the feature. A future WS bridge
  stays a pure relay — DATA/INPUT binary, CTRL text, nothing re-encodes.
- **Replay** (`packages/cli-kit/src/pty-ring.ts`): a ring trimmed to the last
  screen wipe, plus the sticky modes a TUI announces only at startup (mouse,
  bracketed paste, key encoding) — without re-asserting them a late attacher has
  a dead mouse and dead paste.
- **Fan-out**: the CLI, the tray and a plain terminal drive the same agent. Size
  arbitration keys off attach/resize, never keystrokes, so typing in a small
  window can no longer shrink a big one.
- **Lifetime — lease + grace**: a client can lease a session
  (`manager attach --lease-id`, which the tray passes with a persisted id). An
  unpinned session is reaped once every holder stays away past
  `manager.leaseGraceSeconds` (60s): quitting the tray ends the terminals it
  opened, a relaunch or update inside the window takes them back. A session
  nothing ever leased is never reaped; `manager pin` and
  `manager.lifetime: persistent` exempt one or all.
- **tmux stays** for external managers (`send-keys` via `$TMUX_PANE`), the claude
  `--bg` attach session, and as the fallback carrier — `manager.carrier` chooses
  deliberately, and a named carrier never silently falls back.
- **Detection got stronger**: the host exports `AGENTBOX_MANAGER_RUN` and the CLI
  sends it back, so `$AGENTBOX_MANAGER` — which Claude's daemon leaks into
  unrelated sessions — is believed for a pty manager only when the run id matches.

## Verified

`manager start` → `kind: pty`; `manager message` typed **and** submitted into a
live TUI; `hub restart` left the session running and attachable; `manager stop`
ended it and reaped socket + meta; with an 8s grace, a lease holder leaving got
the session reaped, **the same client id returning inside the window kept it**, a
pinned session survived and unpinning reaped it within a tick;
`manager.carrier: tmux` started the same agent on the old carrier. Repo gates:
lint, typecheck and the full suite green.

Bugs the live runs found, each now covered by a test: the spec rode an inherited
fd already at EOF so every start silently fell back to tmux; `stop` authenticated
with an empty token; the hub's sizeless control client won the size arbitration
and `resize(0,0)` **killed the host with the agent inside it**; a deferred
repaint nudge after exit threw `ENOTTY` out of a timer; the pin route answered
the wrong envelope.

## Deploy order

`ManagerRegistration` now carries `kind: 'pty'`, and registration is the one
manager write that travels to a control box — **update the control box and the PC
together**, or a start on the new PC build is refused by the old box. The tray
half landed in `../agentbox-tray` (`453291d`); its live embed is blocked on that
same control-box update.

https://claude.ai/code/session_01X2bfZLcWHE2AHjqFJr5wvY

---

> [!NOTE]
> **Medium Risk**
> Touches manager process lifecycle (spawn, attach, leases, reaping, and cross-machine registration); mistakes could strand or kill live agent sessions, but behavior is heavily tested and tmux remains a fallback.
> 
> **Overview**
> Hub-run **manager** sessions can run on a detached **pty host** (`dist/pty-host.js`) instead of only tmux, so clients get a **raw byte proxy** to the agent terminal (scrollback, mouse, modified Enter stay with the attaching terminal).
> 
> **CLI:** `manager attach` prefers local pty meta, adds `--raw`, `--detach-key`, and `--lease-id` (tray-style leases with TTL refresh), plus **`manager pin`** to keep sessions alive without a lease. Default attach targets any non-`external` running manager, not just tmux.
> 
> **Hub / API:** Managers gain `kind: 'pty'`, `ptyAttach`, and `pinned`; **`POST /managers/{id}/pin`**, pty-aware registration/heartbeat validation, **`runId`** on detect for trustworthy `managerId`, message delivery via **pty inject**, and **carrier-aware** start/resume/stop (pty default / `auto`, tmux fallback) with new **`MANAGER_CARRIER_MISSING`** / **`PTY_CARRIER_MISSING`** errors. A **pty janitor** sweeps dead host leftovers on hub startup.
> 
> **Supporting pieces:** Binary pty wire protocol, replay ring + detach chord (`C-] d`), config keys under **`manager.*`**, OpenAPI/docs updates, and tests for attach, carrier selection, and janitor behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54f4f624b90d96c595a3a413b5141a5bdb0f0ae4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


madarco:
bugbot run

